Issue #29, opened 2018-04-27
Timeout never reached


Tkopic001:
Hello,

I really love your work, this really speeds up auth by a bunch. I have to say, In well lit conditions it gets my face in less then a second so I've put the scan timeout to 2 sec, 

but I've noticed something when i'm in a darker room and there's no way to see the face. It get's stuck and never hits the timeout to offer the password input until I turn on the lights , and then i do it says the timeout has been reached and offers me to input my password. i left it hanging for a few minutes. but nothing happened until i turned on the light and it could see the face. 
I've played a lot with the settings to try to avoid this, but no progress so far,

Maybe you could program a little failsafe if the frame is to dark that it just skips the scan completely? 
i wish i knew python so i could fix this myself and not bother you with this.

Thanks for reading this, 
and keep up the good work!
Cheers!

Tkopic001:
OK, so even though i have absolutely no experience with python code i took the time and read each line of the code you wrote, (Tanks for commenting so much, really helped me to figure it out)
and i've managed to fix this issue for myself, now this is nowhere near a production solution but it did the trick.

I turned on the debugging mode and checked how fast does it really work.
did a few tests with different lighting and it recognizes the face in 1 or max 3 frames.

so just after the loop starts i simply did this in this file:
**/lib/security/howdy/compare.py**
```
...
 # Start the read loop
frames = 0
while True:
        # Increment the frame count every loop
        frames += 1

        **#Fix the timeout issue
        if dark_tries > 10:
                stop(11)**
...
```

And this works pretty well actually. it worked for me, but i hope this helps. 
Have in mind that i have set the dark_threshold to 10.

Tkopic001:
Hello,

I'm glad i could help! :), I work as a SysAdmin so i do have some programing experience with Bash and i can read what PHP code does but i can't write it. I never read or wrote anything in Python. but then again it was really well documented so it was easy to see and understand what each part does. 

I'm using a normal laptop webcam(BisonCam, NB Pro to be precise), it has no IR lights, but Howdy works perfectly fine with it, The only problem I had  were the dark frames, sometimes i use the laptop with the lid closed and connected to the external monitor so the webcam can't see anything and howdy was just stuck waiting so i had to open the lit a little for it to get some light and exit.
 
Now it works like I wanted it to, if there are too many dark frames it means the lid is closed or i'm in the dark so there's no point in trying to find a face and it just exits so that the PAM can ask for a password,  simple.

[...]

Issue #27, opened 2018-04-25
Pink image returned by test?


arifeinberg:
This may be outside the scope, but I'm having trouble adding a model with my IR cam (04f2:b613), and upon using test, I see that the camera is returning a mostly pink image:
![howdy test_screenshot_25 04 2018](https://user-images.githubusercontent.com/1208877/39273155-085b2f0c-48ac-11e8-9614-9adf496ac5ea.png)

I also notice that while it's running, the IR emitters are constantly on, not flashing.   I notice that when adobe flash uses the IR cam (for something else), the IR emitters flash, and a readable grayscale image is produced.  I have no idea where this is controlled.  Is there a parameter specified by the application that calls the camera, or is this determined somewhere else?

arifeinberg:
The Video4Linux preview also returns a usable image.  It seems this camera offers two modes: Motion-JPEG and YUYV raw.  Motion-JPEG is the one that works.  The mode is not set at the level of this software?

Boltgolt:
Could you try replacing line 22 (a white line) in `/lib/security/howdy/cli/test.py` with the following:

```
video_capture.read()
print(video_capture.get(cv2.CAP_PROP_FOURCC))
video_capture.set(cv2.CAP_PROP_FOURCC, 1196444237.0)
```

And see if that fixes anything or throws any errors? I'm not at my PC right now so it might be a little off.

arifeinberg:
I was just a few minutes behind you!  I added the following line after initialization of video_capture (in add, test, and compare):
video_capture.set(cv2.CAP_PROP_FOURCC, 1196444237)
and it is now working (after I raised the dark threshold).

Thanks!

Boltgolt:
Great! I'll add it as a setting in a later release

PR #34, opened 2018-05-10
Version 2.2.1

Boltgolt:
**Changelog:**  
 * Added mechanism to keep config files between updates
  * Added force_mjpeg option to fix YUYV image issues (thanks @arifeinberg!)
  * Revamped the bash autocompletion script
  * Fixed timeout never being reached in certain scenarios (thanks @Tkopic001!)
  * Fixed issue where BGR to RGB frame conversion caused a crash (thanks @Jerezano!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,7 @@ notifications:
   email:
     on_success: never
     on_failure: always
+
+addons:
+  apt:
+    update: true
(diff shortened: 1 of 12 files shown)


Boltgolt:
This includes fixes for #30, #29 and #27

Issue #46, opened 2018-06-20
eliminate subprocess call


dmig:
This
https://github.com/boltgolt/howdy/blob/961342a5222bb1cef2dd6b7c14c46febd1baf7f8/src/cli.py#L13

Can be replaced with
```python
try:
  user = os.getlogin()
except:
  user = os.environ.get('SUDO_USER')
```

boltgolt:
If i recall correctly i tried a load of different solutions for this and my line was the only one that worked in `su`. I'll give yours a spin though, it's much easier to maintain.

Issue #44, opened 2018-06-07
Division by Zero in History, compare.py


stellarpower:
Hi,

I'm getting errors when trying to sudo, evidently the history is missing.

  File "/lib/security/howdy/compare.py", line 101, in <module>
    if float(hist[0]) / hist_total * 100 > float(config.get("video", "dark_threshold")):
ZeroDivisionError: float division by zero
Unknown error: 1

is the backtrace.

I'll try to take a look when I can.

Cheers

boltgolt:
Found the issue: If every single pixel of the frame is exactly rgb(0,0,0), the sum of the histogram will be zero. I'll add a fix in the next patch but for now you can add

```python
	if hist_total == 0:
		dark_tries += 1
		continue
```

on line 100 in `/lib/security/howdy/compare.py`

Issue #41, opened 2018-05-18
Change to use xed instead of gedit?


stellarpower:
I can't use the CLI to open the config as on newer systems gedit's been replaced by xed. IDK if you wanna change this in the source.

Boltgolt:
Not sure, other systems might not support both xed and gedit and i don't think nano or other command line editors are user friendly enough. Might have to look into building an actual UI for the config.

stellarpower:
For now probably testing for the presence of xed and if not calling gedit
or kate would be a good solution, do you think?

dmig:
Solution is simple: use `xdg-open` instead of `gedit` here https://github.com/boltgolt/howdy/blob/84c0501642f761a822d1722f7d4cc395e91c2b6c/src/cli/config.py#L12

dmig:
Or, the better solution: use CLI editor `/etc/alternatives/editor`

boltgolt:
`xdg-open` opens in a wine version of Notepad for me so i'm not sure about that. Both firefox and nautilus open a .ini in gedit by default for me so i'm not sure what `xdg-open` is using here.

`/etc/alternatives/editor` is nice but nano might be a little confusing for less technical people? I'm sure the amount of people who use Ubuntu and not know how to save in nano is very small, maybe a comment in the config file explaining how to close nano would be enough?

Issue #40, opened 2018-05-16
does not pick up ir camera when they're two separate devices


wzrdtales:
In my setup I do have two cameras. One IR on /dev/video0 and the webcam on /dev/video2. From the source and everything else it quite seems like only one of both realy gets used, the one configured. If you configure the webcam, it'll register me just fine. But it does not use the IR at all (does not activate though). If I use the IR it complains about: no face detected.

What is your setup for this? And what are you sending to the library in use here? The IR picture or the webcam picture? Or something combined?

Boltgolt:
Normally during setup you get asked what camera to use by testing them until your emitters turn on. 

The library gets a frame from one of your webcams and returns the faces it detects in that frame. Howdy only sends it the frame from the configured `/dev/videoDEVICE_ID` device (should be 0 in your case). Every working Howdy setup has 2 separate webcams though, this is normal.

I suggest setting your `device_id` to 0 in `sudo howdy config` and running `sudo howdy test` to see what's going wrong. You should see a red circle around your face.

wzrdtales:
@Boltgolt I am on arch, so I just cloned your repo, have had no setup done though.

Have already looked into howdy test. If on device 0 the ir camera picks up a picture (so the green colored one). if on device 2 it picks up the webcam. However on the IR it just does not recognize any faces, but actually if you do not go 2cm before the IR cam you do not really see that much, so that would be expected (you can barely see something by eye but its there though). It would be very hard to find a face in that scenario.

wzrdtales:
```
v4l2-ctl --list-devices --all
Integrated IR Camera: Integrate (usb-0000:00:14.0-5):
	/dev/video0
	/dev/video1

Integrated Camera: Integrated C (usb-0000:00:14.0-8):
	/dev/video2
	/dev/video3

Driver Info (not using libv4l2):
	Driver name   : uvcvideo
	Card type     : Integrated IR Camera: Integrate
	Bus info      : usb-0000:00:14.0-5
	Driver version: 4.16.8
	Capabilities  : 0x84A00001
		Video Capture
		Metadata Capture
		Streaming
		Extended Pix Format
		Device Capabilities
	Device Caps   : 0x04200001
		Video Capture
		Streaming
		Extended Pix Format
Priority: 2
Video input : 0 (Camera 1: ok)
Format Video Capture:
	Width/Height      : 640/480
	Pixel Format      : 'YUYV'
	Field             : None
	Bytes per Line    : 1280
[...]

PR #48, opened 2018-06-28
Version 2.3.0

boltgolt:
**Changelog:**
  * Added a config option to set the frame height and width (thanks @wzrdtales!)
  * Rewrote the code that fetches the non-root username (thanks @dmig!)
  * Changed the config command so it uses the default editor (thanks @stellarpower and @dmig!)
  * Fixed issue where a "y" could be interpreted as a no (thanks @ramkrishna757575!)
  * Fixed division by zeno (thanks @stellarpower!)


boltgolt:
This is a fix for issues #33, #40, #41, #44 and #46

PR #52, opened 2018-07-10
Use variables from config.ini frame_height frame_width for camera res

janecz-n:
Fix camera resolution on `sudo howdy add` to resolve problems link to camera resolution.


boltgolt:
Ah, totally forgot add.py when adding that. Thanks a lot!

PR #64, opened 2018-08-26
add instructions to install python3-opencv

diepes:
After installing repo, i got a error about python not being able to import cv2.

Fixed with apt install python3-opencv

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,11 @@ sudo apt install howdy
 
 This will guide you through the installation.
 
+p.s. ensure you have python3-opencv installed
+```
+sudo apt install python3-opencv
+```
+
 ### Arch Linux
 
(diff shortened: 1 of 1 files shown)


boltgolt:
This is a broader issue than just adding a line in the README unfortunately.

The packaged versions of OpenCV bindings fail on instances of Ubuntu 16.04 (like mine). This means that adding it as a dependency of the package would prevent Howdy from installing at all. To circumvent that issue, Howdy installation includes a full clone and build of dlib and opencv, but on some platforms the python bindings are not fully installed (like Ubuntu 18.04).

So it's currently in a state where 16.04 users can install it without issue, and 18.04 users need to run the apt command. A better solution would be to catch the import error `No module named 'cv2'` and print instructions on installing the apt package, if you're up for it.

PR #103, opened 2018-11-29
Add  Fedora

luyatshimbalanga:


boltgolt:
Looks very good, any build or installation files that could be included in this repo like debian has? (howdy.spec?) I'd like to keep it here locally as much as possible.

luyatshimbalanga:
Only [howdy.spec](https://copr-be.cloud.fedoraproject.org/results/luya/howdy/fedora-rawhide-x86_64/00831058-howdy/howdy.spec) file is needed as build and installation scripts are included. You can put in the folder called for example _rpmspec_ to keep in the git repository.

PR #116, opened 2018-12-23
Adding support for v4l2 devices directly in python

timwelch:
v4l2l.py is included in src dir. The version you get from pip install does not work with python3 (at least not on my system), so I had to modify several of the functions returning a list() of range() instead of the original range().

Since this ties directly into the camera, instead of using ffmpeg, it can actually grab and hold onto the camera instead of reading X frames from the camera at a time. Thus, the camera is opened in a continuously recording mode and a single frame is grabbed each time a request is made to compare a frame. The other added benefit of not using the bloated ffmpeg is that it trims an entire second off of the startup time on my system!

### DIFF
--- a/src/cli/add.py
+++ b/src/cli/add.py
@@ -95,11 +95,13 @@ insert_model = {
 
 # Check if the user explicitly set ffmpeg as recorder
 if config.get("video", "recording_plugin") == "ffmpeg":
-	from ffmpeg_reader import ffmpeg_reader
-
 	# Set the capture source for ffmpeg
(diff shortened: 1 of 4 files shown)


timwelch:
FYI: the v4l2 python has not been updated since 2010, probably why it didn't work on my machine without a few modifications.

https://pypi.org/project/v4l2/

boltgolt:
Thanks a lot! Including the entire library file into howdy isn't the best, but i don't see another option if some of your own modifications are required to make it word.

PR #117, opened 2019-01-02
Added notice that facial detection is occurring

mrkmg:
Added a pam conversation message that the facial detection is being performed.

**Why?**

In my setup, my webcam is not always "front and center". I often use external monitors on my laptop and my webcam ends up pointing toward the side of my face. In order to get face detection to work, I need to turn my head and look into the webcam.

I find myself forgetting to turn and look at the webcam when typing sudo from the command line. Then I either type my password or canceling it and trying again.

I added this message on my machine, and it works great to remind me to look at the webcam.

### DIFF
--- a/src/config.ini
+++ b/src/config.ini
@@ -2,6 +2,9 @@
 # Press CTRL + X to save in the nano editor
 
 [core]
+# Print that face detection is being attempted
+show_detection_attempt = false
+
 # Do not print anything when a face verification succeeds
 no_confirmation = false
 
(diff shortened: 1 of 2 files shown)


boltgolt:
Thanks for adding your work back into Howdy! 
Because this is new behavior that all users will see after the update to 2.5.0, i would rather have that it could be turned on or off with a config value. Would you be comfortable with adding a value to the [core] section of the config file?

mrkmg:
I assume we want this to be config off by default?

boltgolt:
Thanks a lot!

Issue #122, opened 2019-01-11
Update AUR (Possible addition of -git package?)


joeknock90:
The AUR PKGBUILD is out of date for 2.5.0

I've attached a working PKGBUILD for 2.5.0 if you want to use it. 

Might I also suggest a howdy-git package to build from github directly from the AUR?

Thanks!

[PKGBUILD-2.5.0.txt](https://github.com/boltgolt/howdy/files/2749768/PKGBUILD-2.5.0.txt)

boltgolt:
Thanks for providing your modified PKGBUILD! Some major things have changed in Howdy between 2.3.1 and 2.5.0, notably including removing face-recognition as an dependency and requiring extra optimization steps for dlib compilation. I currently don't see these changes in the PKGBUILD and i'm hesitant to push a working but incomplete script at the moment.

boltgolt:
If you want to, you could port [the changes](https://github.com/boltgolt/howdy/compare/v2.3.1...master#diff-07e2cd25f7da0550504d107fb2cae5be) (press "Files changed") from the debian installer to the PKGBUILD. The wiki is also out of date for 2.5.0 i think.

boltgolt:
Let me know if you have any issues figuring out what the postinst is doing

depadiernos:
Just wondering if this postinst python script can just be use as-is as the post_install function mentioned here: https://git.archlinux.org/pacman.git/plain/proto/proto.install. :thinking:

boltgolt:
Sounds good, and might also be able to automate the [uninstallation steps](https://github.com/boltgolt/howdy/blob/master/debian/prerm).

depadiernos:
Looks like somebody updated the AUR pkgbuild, but it's not working. :thinking: Getting:
```
127 dennis@zelda2 /usr/lib/security % sudo howdy add                          :(
Data files have not been downloaded, please run the following commands:

	cd /usr/lib/security/howdy/dlib-data
	sudo ./install.sh
```
But there is no install.sh in that folder.

boltgolt:
That folder should include a [readme](https://github.com/boltgolt/howdy/tree/master/src/dlib-data) as well with commands to run manually. It's something the something the PKGBUILD should do automatically in my opinion.

depadiernos:
Ah, I see!

pthalamy:
I am running into new issues updating howdy from AUR in its current state. I was previously running howdy 2.3.1 without issue.

The installation fails during the `package()` phase.

[...]

PR #131, opened 2019-01-21
Added notification popup when it's running

Ev1lbl0w:
Added a notification/popup *(through `notify-send`)* to notify the user when the app is scanning faces.

The message, timeout and icon of the notification are examples, so they're obviously open for change.

I also made this feature optional by adding it to `config.ini`*(by default it's active: again, open for change)*

### DIFF
--- a/src/config.ini
+++ b/src/config.ini
@@ -29,6 +29,9 @@ disabled = false
 # computational power to run, and is meant to be executed on a GPU to attain reasonable speed.
 use_cnn = false
 
+# Show a notification to the user when face detection is being attempted
+show_notification = true
+
 [video]
 # The certainty of the detected face belonging to the user of the account
 # On a scale from 1 to 10, values above 5 are not recommended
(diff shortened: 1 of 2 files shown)


zeka001:
I don't think this feature is useful, why I will like to have a notification when the app is running? I know how my system works and what I have install. Anyway, at least don't make it active by default, I will vote for that.

reverland:
In fact I always feel the system freezes at the beginning, before I realize its howdy hardly trying to recognize me.

boltgolt:
First of all, thanks a lot for immediately making a PR for something you'd like to see in Howdy! I've been working on a GUI very much like the one [fingerprint-gui has](http://www.ullrich-online.cc/fingerprint/doc/test.png) while authenticating, but this would help a lot in the meantime. 

I'm curious though, what does this option do for you guys that the `detection_notice` config option does not?

Ev1lbl0w:
> I'm curious though, what does this option do for you guys that the `detection_notice` config option does not?

Uh. The AUR package seems to be outdated, because I didn't realize something similar was already implemented. My bad :wink: 

So I think the only difference here is that one is a popup, and the other is a text message. But I actually believe your solution is much better, since it works in situations `notify-send` doesn't *(like lightdm)*.

Feel free to close this PR, I now noticed this is redundant.

boltgolt:
The arch package is behind 2 version compared to the Debian and Fedora releases, it's being tracked in #122. I'm gonna close this PR then, but once again thanks a lot for merging your changes back into Howdy!

PR #155, opened 2019-02-25
Add video_format options to opencv backend

AnthonyWharton:
Also fixes a bug with a GStreamer-CRITICAL error message appeared every time the `video_capture` was opened.

As an aside, there seems to be a lot of code duplication in setting up the `video_capture` objects, I may try clean that up into a utils file or something to that efffect, but I am not familiar enough to start refactoring the project! Let me know what you think :smiley:

----------------

Example of the bug mentioned above:
```
# G_DEBUG=fatal_warnings python compare.py username                                                                                       
```
Output:
```
(python:16551): GStreamer-CRITICAL **: 18:06:31.199: gst_element_get_state: assertion 'GST_IS_ELEMENT (element)' failed
[1]    16551 trace trap (core dumped)  G_DEBUG=fatal_warnings python -W error compare.py username
```
*(Note that the `G_DEBUG` environment variable in this example is there simply to crash the program when the error occurs. This message did not actually seem to stop execution in practise, but having a warning pop up all the time is a little bit irritating!)*

### DIFF
--- a/src/cli/add.py
+++ b/src/cli/add.py
@@ -108,8 +108,16 @@ elif config.get("video", "recording_plugin") == "pyv4l2":
 	from recorders.pyv4l2_reader import pyv4l2_reader
 	video_capture = pyv4l2_reader(config.get("video", "device_path"), config.get("video", "device_format"))
 else:
+	# Lookup dictionary for different format strings in the config file -> opencv enum
+	opencv_api_pref = {
+		"v4l2"   : cv2.CAP_V4L2,
+		"vfwcap" : cv2.CAP_VFW
+	}
 	# Start video capture on the IR camera through OpenCV
-	video_capture = cv2.VideoCapture(config.get("video", "device_path"))
+	video_capture = cv2.VideoCapture(
+			config.get("video", "device_path"),
+			opencv_api_pref.get(config.get("video", "device_format"), cv2.CAP_V4L2)
+	)
 
 # Force MJPEG decoding if true
 if config.getboolean("video", "force_mjpeg", fallback=False):
--- a/src/cli/test.py
+++ b/src/cli/test.py
@@ -20,8 +20,16 @@ if config.get("video", "recording_plugin") != "opencv":
 	print("Aborting")
 	sys.exit(12)
 
-# Start capturing from the configured webcam
-video_capture = cv2.VideoCapture(config.get("video", "device_path"))
+# Lookup dictionary for different format strings in the config file -> opencv enum
+opencv_api_pref = {
+	"v4l2"   : cv2.CAP_V4L2,
+	"vfwcap" : cv2.CAP_VFW
+}
+# Start video capture on the IR camera through OpenCV
+video_capture = cv2.VideoCapture(
+		config.get("video", "device_path"),
+		opencv_api_pref.get(config.get("video", "device_format"), cv2.CAP_V4L2)
+)
 
 # Force MJPEG decoding if true
 if config.getboolean("video", "force_mjpeg", fallback=False):
--- a/src/compare.py
+++ b/src/compare.py
@@ -123,8 +123,16 @@ elif config.get("video", "recording_plugin") == "pyv4l2":
 	from recorders.pyv4l2_reader import pyv4l2_reader
 	video_capture = pyv4l2_reader(config.get("video", "device_path"), config.get("video", "device_format"))
 else:
+	# Lookup dictionary for different format strings in the config file -> opencv enum
+	opencv_api_pref = {
+		"v4l2"   : cv2.CAP_V4L2,
+		"vfwcap" : cv2.CAP_VFW
+	}
 	# Start video capture on the IR camera through OpenCV
-	video_capture = cv2.VideoCapture(config.get("video", "device_path"))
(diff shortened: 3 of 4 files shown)


AnthonyWharton:
Whoops - let me have a look at what tripped the test as everything seemed to work on my machine.

boltgolt:
First of all, thanks for contributing! The changes look good, but the failing test is a bummer. I wasn't aware some installations were throwing these (semi?)critical errors so a fix is definitely welcome.

As for the Don't Repeat Yourself part, i had the same thought. The initialization code started small enough but has steadily grown each release and really needs its own file now. It's on the to-do list.

AnthonyWharton:
No problem at all, happy to help! Sorry I disappeared for a bit, going to have a closer look now.

Not entirely sure why the critical error appears as it still works fine! Just a nuisance really.. Looking around online, there are various other people citing that this error/similar ones seem to be due to missing packages that the back-end(s) require. I have all these installed and thus I imagine this is why it works fine. In my searching I found [one stack overflow answer](https://stackoverflow.com/a/53543167) that suggested the fixes that I added as a workaround in the patch (namely moving away from GStreamer). I also took that moment to add the extra codec support as opencv seemed to support it.

Back to the test failing issue issue, from an initial glance it appears that the OpenCV API's don't work with the videos in the test file only from my testing. The reason I committed and PR'd was because the patch works fine with my IR camera on my machine (a Dell XPS 13 9370). Going to dig deeper now, unless you have any ideas!

AnthonyWharton:
Ok, so anyone can correct me if I'm wrong, but it looks like the video for Linux drivers (`v4l`/`v4l2`) APIs only support reading from a camera device and not a file. Thus as this change uses these APIs it breaks the test!

At least on my install of `opencv` (which was fairly stock from the arch repositories) the only API's that support reading from files are `GStreamer` (which gives the warning printout), and `ffmpeg` (which seems to defeat the purpose of using `opencv` as you have a dedicated `ffmpeg` backend already).

As an extra test, I also tried changing the config to use `ffmpeg` and ran the tests locally, and sure enough the `ffmpeg` backend using `v4l2` breaks the test script as well. This arguably could be seen as an issue that could be raised separately later...

There's a lot of options to remedy this which I'd be happy to help with, but I'd like some of your input if possible. A couple of options I see are:

1) Modify the tests to use something like [v4l2loopback](https://github.com/umlaeute/v4l2loopback/) to emulate a camera device (some example commands on the wiki of the linked repo.) - maybe that could work better with `opencv`/`ffmpeg` set to use `v4lc`. There's also room here to make the tests slightly more robust as they only test the default config currently.
2) Introduce an option into the config to select the `opencv` [backend API](https://docs.opencv.org/4.0.1/d4/d15/group__videoio__flags__base.html#ga023786be1ee68a9105bf2e48c700294d), and set the default to something that passes the test script. This obviously leaves cases that are known to fail the test scripts used by travis, but gives users the option to change backends (and in my case suppress the critical error message!). 

The first option will be a bit more work, especially integrating with Travis (which I'd need some help with!). However this obviously makes the tests more thorough for the future. The second option I could implement for you pretty quickly but just feels like I am playing a box ticking exercise with the test suite - although arguably this is already happening as the `ffmpeg` backend breaks the tests currently :stuck_out_tongue: 

Don't mean to poke holes in the project, it's super cool and I'm just trying to help make things better :smiley: I'll leave it as your call as to how things proceed!

boltgolt:
I've looked into v4l2loopback when i was first writing the tests, but building and running it added about a minute to test time while it was just easier to read directly from a video file. It seems we've hit its limitations though.

I think the way forward from here is to ditch the way we read video files right now and go back to compiling and installing v4l2loopback. I'm no expert at Travis either and i barely have the time to comment on PRs and issues right now so i will not be working on it for some time. Feel free to try it if you want to, any pushes you make to your master branch will also trigger a restart of the Travis test.

Anyway, thanks for the excellent write-up! Projects like Howdy really only evolve because people like you take the time and effort to push them further.

AnthonyWharton:
Sweet! I can't make any promises as to when I will be able to add more to this PR, as in theory I also have similarly little free time :stuck_out_tongue: That being said, I have a habit of procrastinating from work I'm meant to be doing with projects like this, so I'll have a go at upgrading the tests a bit at some point.

Like yourself, I'm not an expert at Travis, so we'll work that out as we go! I'll post any more major updates/design decision queries here as and when I have them.

AnthonyWharton:
Quick update: `v4l2loopback` is not really an option as Travis will not support kernel modules which are required. Will have a little re-think!

AnthonyWharton:
I really shouldn't have done this from my fork's master branch.. Was just bringing it inline with upstream with a rebase. Any force push in future will be for that unless otherwise stated.

boltgolt:
I appreciate your work on this a lot, so don't take that moved milestone as pressure or anything.

AnthonyWharton:
No worries.. I've had to leave this as I'm sorry busy with my degree.. I'll get around to this probably in the summer!

AnthonyWharton:
Closing this PR to move it into another branch. Will reopen the PR from the new branch.
Should never have been from my `master` branch!

PR #156, opened 2019-02-27
Change Debian to Ubuntu

rileyball:
Debian can't use PPAs so saying that you install this on debian like that is misleading

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Howdy is currently available for Debian/Ubuntu, Arch Linux and Fedora. If you’
 
 **Note:** The build of dlib can hang on 100% for over a minute, give it time.
 
-### Ubuntu (Ubuntu, Linux Mint, etc)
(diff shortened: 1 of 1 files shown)


boltgolt:
Seems a bit weird to say "Ubuntu" and then immediately list Ubuntu as the first example. What would you think of linking to the .deb on the [releases](https://github.com/boltgolt/howdy/releases) page (which should work on Debian)?

rileyball:
Edited it so debian is now in a separate category with the releases page linked

boltgolt:
Thanks a lot!

CybernetiX-S3C:
Debian actually does support PPAs. LMAO

boltgolt:
It does not support them out of the box though. [see this comment](https://github.com/boltgolt/howdy/issues/162#issuecomment-468923354)

Issue #45, opened 2018-06-19
Running howdy on Arch Linux


KelleyMcChes:
Hey boltgolt,

Earlier, I was curious about getting howdy to run on Fedora and tried to get it running. After it initially didn't work, I just gave up on it. Now that I am on Arch Linux, I retried installing it and finally got it to work with howdy 1.0.0. I don't know how to convert the current deb packages to files that pacman uses, but if you do have an idea, I can give you the current (working) installer.py file. There is no common-auth file in Arch Linux, so you would need to enter the pam.d information into each pam config file.

boltgolt:
Thanks for taking the time to look at this! The latest version should probably still be usable if you copy the contents of the [src](https://github.com/boltgolt/howdy/tree/master/src) folder. I'm looking into packaging for arch and fedora when i have the time so please do share your installation script. 

Could i @ mention you if i have a pacman version to test?

KelleyMcChes:
After copying all of the files from the current src folder into /lib/security/howdy and manually symlinking the cli.py file into /usr/bin/howdy, the current howdy works. I had to change the /dev/video* to the correct one with `sudo howdy config` but it is perfect.

KelleyMcChes:
If you would like me to, I can try packaging it for you

boltgolt:
That would be great. Does arch use a separate folder (like `/debian` on apt systems) to store the packaging files? I'm reading the packaging guide and i don't fancy putting the fedora and arch files directly in the root. 

The `/debian/postinst` file contains the current installation script, might be worth looking at.

If you do get a working build, a pull request with the source would be amazing

KelleyMcChes:
Well since howdy uses packages from the AUR, your only choice is using the AUR. This will be much easier since you can basically use any format for your source. How the AUR works is there is a file called the PKGBUILD that automates the build and install process. The package is installed in a fakeroot environment and then put into a tar.xz file (Arch's packaging style) and automatically installed. Ignoring a few minor modifications you might need to make when designing the PKGBUILD, it is basically developing a set of instructions to install from source.

boltgolt:
Sounds easier than debian, must the PKGBUILD be placed in the root or can the package be build from a subfolder, like `/arch`?

KelleyMcChes:
The file is actually stored in git server on the AUR while your files just need to stay in GitHub. [Here](https://aur.archlinux.org/packages/firefox-beta/) is an example of an AUR package.

KelleyMcChes:
[...]

PR #161, opened 2019-03-01
Add Arch Linux PKGBUILD

AnthonyWharton:
As the `.deb` files are in the repo I though I'd add an updated Arch Linux PKGBUILD file. 

**This also needs to be updated on the AUR as it's very out of date!** I'm happy to do this if you don't want to maintain things over there, but it would be nice to be updated :smile:

### DIFF
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -1,7 +1,7 @@
 # Maintainer: boltgolt <boltgolt@gmail.com>
 # Maintainer: Kelley McChesney <kelley@kelleymcchesney.us>
 pkgname=howdy
-pkgver=2.3.1
+pkgver=2.5.0
 pkgrel=1
 pkgdesc="Windows Hello for Linux"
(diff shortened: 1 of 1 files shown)


boltgolt:
Thanks a lot, I've [wanted this for a long time](https://github.com/boltgolt/howdy/issues/45#issuecomment-398679286) but I'm very hands-off when it comes to arch because i simply can't test anything i do. 

This PKGBUILD is missing a lot of dlib buildtime optimizations that the debian script has, but I'm uploading it anyway because the PKGBUILD we have now is all the way back from 2.3.1.

AnthonyWharton:
Sure, it was the bare minimum to get it up and running as it's a little bit behind!

I'll take a look at the optimisations when I get some time and see if I can add those in. Are you going to update the `PKGBUILD` on the AUR?

boltgolt:
Pushing it as we speak.

PR #166, opened 2019-03-10
Manual exposure support for OpenCV

accek:


src/compare.py
```diff
@@ -144,6 +144,17 @@ def stop(status):
...
 video_capture.grab()
 
+# Disable autoexposure
+exposure = config.getint("video", "exposure", fallback=-1)
+if exposure != -1:
+	# For a strange reason on some cameras (e.g. Lenoxo X1E)
+	# setting manual exposure works only after a couple frames
+	# are captured.
+	for i in range(5):
+		video_capture.grab()
```
boltgolt: I don't really like this delay. Could you possibly use the main loop, which also counts frames, to check and run this?

accek:
In testing I noticed that the previous method of waiting a couple frames was still not fully reliable. Now I set manual exposure at every frame in the main loop.

Is it ok for you? Or should I rather make this workaround separately configurable or somehow whitelisted per-camera-model?

boltgolt:
Thanks a bunch!

PR #167, opened 2019-03-10
Disable Howdy if lid is closed

accek:


boltgolt:
Thanks for the PR!

The config value name "ignore_closed_lid" kind of implies that Howdy will ignore the closing of the lid and fire anyway, while in fact it will do the opposite. This is also a problem with the "ignore_ssh" value i wrote so i'll rename both at some point.

Is there any way to do this without importing glob?

accek:
> Thanks for the PR!
> 
> The config value name "ignore_closed_lid" kind of implies that Howdy will ignore the closing of the lid and fire anyway, while in fact it will do the opposite. This is also a problem with the "ignore_ssh" value i wrote so i'll rename both at some point.

No problem with renaming. For me the meaning of ignore_ssh was clear.

> Is there any way to do this without importing glob?

Why without importing glob? You mean performance reasons? I don't think this one is noticeable.
Other approaches may include:

- assuming it's just /proc/acpi/button/lid/LID/state, as it is on my machine, but I'm not sure what it the source of the "LID" name and I prefer being safe

- make it configurable, but this is bothersome for the user

- reimplement glob using os.listdir

boltgolt:
I did not know that `LID` could have different names depending on the hardware. It's `LID0` on my machine so using glob was the right choice. Thanks for your contributions!

PR #190, opened 2019-04-13
Add Fedora spec file and polkit rule

luyatshimbalanga:


boltgolt:
Thank so much for your continued efforts on this!

PR #209, opened 2019-05-14
Expanded Python versions in .travis.yml

krichter722:


boltgolt:
I used to test against [more python versions](https://github.com/boltgolt/howdy/commit/c6a7e06d5a534c6172d9864bc4a193f46c989e7b#diff-354f30a63fb0907d4ad57269548329e3L3), but it slightly slowed down the already long tests and never found incompatibilities between python versions. Is it worth it to run so many slow (~10min) tests?

AnthonyWharton:
Not that my knowledge of python version changes is that great, I would say it might be sensible to keep Python 2.7, but then only the latest stable version on Python 3.X?

krichter722:
> Not that my knowledge of python version changes is that great, I would say it might be sensible to keep Python 2.7, but then only the latest stable version on Python 3.X?

Mine neither. For starters, I assume they'll not keep more version branches than now even in the far future. I've seen bugs with Python software only occurring in the latest, only in the oldest or in any imaginable range of versions, especially when it comes to installation routines involving building software from source. Some where caused by differences between OS versions, but some exclusively by the Python version difference.

Since the build already takes 10 minutes you probably have few workflows where you wait for the CI, so the increase isn't that severe in my opinion. Another approach could be: Close this PR and keep the branch and wait until you get into a situation where a bug is reported which could have been avoided in development if you had more Python versions checked in CI.

boltgolt:
Python 2.7 isn't supported, it's not compatible with imports like `configparser` (was `ConfigParser` in python2), and using print as a function instead of a statement.

Would it be an acceptable compromise to just check against the oldest python3 version (3.4), most popular version (3.6) and the latest development version (3.8-dev)?

krichter722:
@boltgolt I suggest adding 3.7 as well because it's used in popular systems like Debian and Ubuntu, see updated commit.

boltgolt:
Thanks a lot!

PR #213, opened 2019-05-18
Fixed build issue

rhysperry111:
Could't install howdy yesterday. Turns out the --yes option has been removed. Removing --yes from pkgbuild allowed me to install


boltgolt:
Thanks a lot!

PR #224, opened 2019-06-27
Update PKGBUILD

debanjanbasu:
edited the version number, source url to new version and updated the sha sum for the new version. hopefully this is all that is needed to be edited.

tested and worked for me


boltgolt:
Thanks! Will push to AUR asap

PR #227, opened 2019-07-05
Debian: Use fswebcam Instead of Streamer

MisterRager:
Buster doesn't seem to have `streamer` in the default repositories, so I refactored the config scripts to use `ffmpeg` instead. It's working on my laptop in a Gnome 3 setup.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+# These are supported funding model platforms
+custom: https://www.buymeacoffee.com/boltgolt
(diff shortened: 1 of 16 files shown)


MisterRager:
So, problem in CI: this only builds if `ffmpeg` is available, but Ubuntu Trusty (Travis' Ubuntu version) uses `libav` instead. I'll see what it'll take to make this work for both.

boltgolt:
Thank you for taking the time to make and test this! Howdy used to use `fswebcam` (even [the comment above the streamer call](https://github.com/boltgolt/howdy/blob/master/debian/preinst#L75) still mentions fswebcam) but switched to streamer because it was installed by default on Ubuntu. Seems like we'll be switching back now, at least on Debian.

PR #258, opened 2019-09-28
Fix small typo in compare.py

Ajayneethikannan:


boltgolt:
Thanks! The variable used to be called "timout" in the whole program before, it was [renamed](https://github.com/boltgolt/howdy/blob/master/debian/postinst#L70) in 2.5.0, but looks like i forgot about this instance

PR #266, opened 2019-10-17
Small README.md fix

Schnilz:
openSUSE was left out ... and technicaly howdy is available to all linux distributions but only packaged for some ;)


boltgolt:
Thanks!

PR #274, opened 2019-11-01
Fix missing colors in overlay

PetePriority:
When running `howdy test` the overlay is shown in grayscale as `overlay` is a copy of `frame` which was already converted to grayscale.

This PR converts grayscale back to BGR after applying Contrast Limited Adaptive Histogram Equalization, and `frame` back to BGR before applying the overlay.


boltgolt:
Good catch, thanks!

PR #276, opened 2019-11-01
Apply CLAHE to frames in add.py and compare.py

### LINKED ISSUE #275: Different behavior in classification of dark frames in add.py/compare.py vs test.py
When adjusting the `dark_threshold` parameter then verification with `test.py` leads to different results than `add.py` or `compare.py`. For instance, frames that are valid in `test.py` might be considered too dark in `add.py`/`compare.py`

----

I've searched for similar issues already, and my issue has not been reported yet.

Howdy version: Commit b49bde1

PetePriority:
Contrast Limited Adaptive History Equalization is not applied in `add.py` and `compare.py` before determining whether a frame is too dark. This leads to different behavior in classifying frame-darkness than what happens in `test.py`.

This should fix #275.

### DIFF
--- a/src/cli/add.py
+++ b/src/cli/add.py
@@ -139,12 +139,15 @@ enc = []
 frames = 0
 dark_threshold = config.getfloat("video", "dark_threshold")
 
+clahe = cv2.createCLAHE(clipLimit=2.0, tileGridSize=(8, 8))
+
 # Loop through frames till we hit a timeout
 while frames < 60:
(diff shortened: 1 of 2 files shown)


DavidM42:
I tested this and it worked for me to fix #269 used together with [this](https://github.com/boltgolt/howdy/issues/269#issuecomment-548105971)

boltgolt:
Thanks a lot!

Issue #233, opened 2019-07-14
howdy does not work in Fedora 30


shatadru:
_Please describe the issue in as much detail as possible, including any errors and traces_

The installation works fine. After I set device_path=/dev/video0 it allows me to do "howdy add"
however, it is not activated when I try to login.

Linux distribution (if applicable):
Fedora Workstation 30

Howdy version:
 rpm -q howdy
howdy-2.5.1-2.fc30.x86_64

Fedora 30 Cinnamon desktop. 

I suspect the issue is related to pam config.

iamzubin:
can confirm this issue.

dill-wood:
Can confirm this is an issue for me as well. Program seems to install correctly. No errors displayed. Can successfully run 'sudo howdy test', but program does not seem to be registering active for the login screen and the terminal prompt.

iamzubin:
did anyone get a workaround?

dill-wood:
Not so far. Was seeing if @boltgolt might have an idea when he had the chance. Wasnt in a huge rush :)

Hakun1n:
Same for Fedora29 (i did not upgraded yet). Test works but no camera/IR gets activated (no red squeres in IR emmiters) during sudo nor logon. 

Thank you in advance for fixing this!

Alex

alfarock:
Can confirm. 

Fedora 30
Dell XPS 9370 

"sudo howdy test" works on X11 but I get the following:

[ INFO:0] VIDEOIO: Enabled backends(7, sorted by priority): GSTREAMER(1000); V4L2(990); OPENNI(980); CV_IMAGES(970); CV_MJPEG(960); FIREWIRE(950); UNICAP(940)
OpenCV | GStreamer warning: GStreamer: unable to start pipeline
 (/builddir/build/BUILD/opencv-3.4.4/modules/videoio/src/cap_gstreamer.cpp:872)

(python3:7282): GStreamer-CRITICAL **: 21:30:52.997: gst_element_get_state: assertion 'GST_IS_ELEMENT (element)' failed
OpenCV | GStreamer warning: GStreamer: unable to query pipeline state (/builddir/build/BUILD/opencv-3.4.4/modules/videoio/src/cap_gstreamer.cpp:420)



Had to add "/dev/video2" to the config.ini file to get "howdy test" working.  I also tried adding "auth sufficient pam_python.so /lib64/security/howdy/pam.py" to the files in /etc/pam.d  but that didn't work.

[...]

PR #321, opened 2020-03-01
Fix the Fedora .spec file

MusiKid:
The dependency specified in the .spec file is not the good one ([#233](https://github.com/boltgolt/howdy/issues/233#issuecomment-557154139), thanks @DnsZhou). `pam_python` requires to install the RPM Sphere repo.


boltgolt:
Thanks! I'll push it to copr ASAP

MusiKid:
No problem!

PR #346, opened 2020-05-02
Fix fedora dependencies

principis:
- howdy uses by default opencv, so v4l2 isn't necessary.
- `python3-face_recognition` and `python3-face_recognition_models` aren't used any more.

v4l2 could be added as suggestion, but then ffmpeg and python-ffmpeg should be added as well.

Edit: missed the dev branch, rebased.


boltgolt:
Thanks for the changes and doing the rebase!

PR #366, opened 2020-06-09
Add darkness threshold feedback

andrewmv:
Allow Howdy to meaningfully distinguish between a registration or compare that failed due to insufficient lighting as opposed to bad camera reads or a facial recognition miss. Report this distinction to the user.

I found that properly tuning the dark_threshold config variable, or even knowing that it *needs* to be tuned, was exceedingly difficult, and this change provides the user with the feedback necessary to adjust dark_threshold appropriately.

In particular, on my Dell XPS 13 9300 integrated Windows Hello camera, I found that 100% of frames were being rejected as "too dark" before even being considered for face detection under the default settings, unless I situated myself in front of a wall. These changes assisted me in finding a working value for dark_threshold.

### DIFF
--- a/src/cli/add.py
+++ b/src/cli/add.py
@@ -135,12 +135,22 @@ time.sleep(2)
 
 # Will contain found face encodings
 enc = []
-# Count the amount or read frames
+# Count the number of read frames
 frames = 0
+# Count the number of illuminated read frames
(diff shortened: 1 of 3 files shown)


boltgolt:
Thanks a bunch!

PR #379, opened 2020-07-01
Update Archlinux PKGBUILD to v2.6.0

kageurufu:


boltgolt:
Thanks for your continued work on the arch build! I appreciate it a lot

PR #381, opened 2020-07-01
Use cv2.CAP_V4L to silence warnings

### LINKED ISSUE #318: Gstreamer warnings
_Please describe the issue in as much detail as possible, including any errors and traces_

While attempting to use howdy, it gives me these warnings:
```
[ WARN:0] global /build/opencv/src/opencv-4.2.0/modules/videoio/src/cap_gstreamer.cpp (1758) handleMessage OpenCV | GStreamer warning: Embedded video playback halted; module source reported: Could not read from resource.
[ WARN:0] global /build/opencv/src/opencv-4.2.0/modules/videoio/src/cap_gstreamer.cpp (888) open OpenCV | GStreamer warning: unable to start pipeline
[ WARN:0] global /build/opencv/src/opencv-4.2.0/modules/videoio/src/cap_gstreamer.cpp (480) isPipelinePlaying OpenCV | GStreamer warning: GStreamer: pipeline have not been created
```

----

Linux distribution (if applicable): Arch Linux 86_64

Howdy version: 2.5.1-2

kageurufu:
Fixes #318


boltgolt:
This seems to make Travis fail, is this an issue with the test or with the code?

principis:
I think it should be `cv2.CAP_V4L2` (atleast that's what I use) but not sure.

kageurufu:
Its a testing issue, since the `sed` command doesn't strip the `cv2.CAP_V4L` part, which then causes it to error because the test video isn't a V4L device

@principis They're the same thing, and the cv2 docs have `CAP_V4L` as the "primary", https://docs.opencv.org/3.4/d4/d15/group__videoio__flags__base.html#gga023786be1ee68a9105bf2e48c700294da78d2234242b9be9c53196be7c2bb2537

towo:
FWIW, I can confirm just hotpatching the file with this change fixes the issue and doesn't seem to have side effects.

boltgolt:
PR seems to be outdated after this long, but doesn't seem to break anything in my install either, implemented in b04ffe5

PR #383, opened 2020-07-02
Fixed typo that caused an error

h45h74x:


boltgolt:
Thanks for catching this!

PR #395, opened 2020-07-16
Update howdy.spec for OpenSUSE

KaratekHD:
I experienced some problems with this .spec file, throwing errors about non existing files. These changes fixed them.
However, installing this RPM is still not possible, since ./install.sh is not existing.


boltgolt:
Pinging @dmafanasyev, is this a good change for the opensuse version?

boltgolt:
No response so assuming it's good!

Issue #285, opened 2019-11-16
Error with Howdy test


Raymo111:
_Please describe the issue in as much detail as possible, including any errors and traces_
Error on 'howdy test':
```
 howdy test
[ WARN:0] global /build/opencv/src/opencv-4.1.2/modules/videoio/src/cap_gstreamer.cpp (1756) handleMessage OpenCV | GStreamer warning: Embedded video playback halted; module source reported: Could not read from resource.
[ WARN:0] global /build/opencv/src/opencv-4.1.2/modules/videoio/src/cap_gstreamer.cpp (886) open OpenCV | GStreamer warning: unable to start pipeline
[ WARN:0] global /build/opencv/src/opencv-4.1.2/modules/videoio/src/cap_gstreamer.cpp (480) isPipelinePlaying OpenCV | GStreamer warning: GStreamer: pipeline have not been created

Opening a window with a test feed

Press ctrl+C in this terminal to quit
Click on the image to enable or disable slow mode

QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
Traceback (most recent call last):
  File "/usr/bin/howdy", line 101, in <module>
    import cli.test
  File "/usr/lib/security/howdy/cli/test.py", line 206, in <module>
    time.sleep(.5 - frame_time)
ValueError: sleep length must be non-negative
```

Also, my cpu shoots to 100%, and it does not work when I log in (i3lock, sddm, etc.)

Linux distribution (if applicable):
Arch
Howdy version:
2.5.1-2

boltgolt:
Impressive, that means you're using more than 500ms to process a single frame. Could you try to check which call between line 111 and 202 in test.py takes so long?

Raymo111:
Err... how would I go about doing that?

boltgolt:
I'm not sure how good you are with python, but you can insert the following line and move it down a line every time:

```
print(time.time() - frame_tm)
```

This will print the execution time since line 111. Make sure you use the right indentation

Raymo111:
Ooh thanks! I'll try that.

PR #401, opened 2020-07-25
Fixes error with slow mode on low FPS

willwill2will54:
When frame_time exceeds 0.5 this change caps the sleep time to be at least zero, as trying to sleep a negative duration crashes the program.

Fixes, or at least is a work-around for #285


boltgolt:
Good catch, thanks!

Issue #385, opened 2020-07-04
Version 2.6.0 "Broadcast message from systemd-journald" problem on Manjaro


bobzhu418:
After upgrading to version 2.6.0, every time if I use sudo on terminal, I get following message. It's not fatal but very irritating. 

```
Broadcast message from systemd-journald@manjaro (Sat 2020-07-04 10:08:23 +08):

sudo[2580]: [HOWDY] Attempting facial authentication for user bob


Broadcast message from systemd-journald@manjaro (Sat 2020-07-04 10:08:25 +08):

sudo[2580]: [HOWDY] Login approved
```

Linux distribution (if applicable): Manjaro 20.0.3 Kernel 5.7.0-3

Howdy version: 2.6.0

bobzhu418:
Ok, I found a work-around. Set `ForwardToWall=no` in `/etc/systemd/journald.conf` to disable the broadcasting.

GarvTambi:
Sometimes face detect successfully and sometimes it shows Unknown Error:1 Why?
Please help!

![dfg](https://user-images.githubusercontent.com/42083679/87365253-609d3000-c593-11ea-9e8f-707f7e887fec.jpeg)

boltgolt:
Thanks for reporting! To keep discussion about it in one place i'm going to close this now, see #374

Issue #374, opened 2020-06-24
Stop using broadcast / journald to return messages


jmlgo:
_Please describe the issue in as much detail as possible, including any errors and traces_

Since the last update, every time howdy is attemping an authentication, I get notifications (sound, system popup) and ALL my terminals get messed with a broadcast message, even when `detection_notice = false` is configured.

```
Broadcast message from systemd-journald@xxxxxx (Wed 2020-06-24 12:07:18 CEST):

sudo[11610]: [HOWDY] Attempting facial authentication for user xxxxxx


Broadcast message from systemd-journald@xxxxxx (Wed 2020-06-24 12:07:20 CEST):

sudo[11610]: [HOWDY] Login approved
```

Pretty similar to this: https://github.com/s3fs-fuse/s3fs-fuse/issues/301

----

I've searched for similar issues already, and my issue has not been reported yet.

Linux distribution (if applicable): Kubuntu 20.04

Kernel: 5.4.0-37-generic

Howdy version: Howdy 2.6.0

principis:
The syslog api is used incorrectly.

For example https://github.com/boltgolt/howdy/blob/master/src/pam.py#L38
```python
syslog.syslog(syslog.LOG_AUTH, "[HOWDY] Attempting facial authentication for user " + pamh.get_user())
```

The first parameter is the priority, but the `LOG_AUTH` constant is passed, which is used for the facility causing the message to be logged with emerg severity.
It should be changed to something like this, with appropriate `priority` for each message.

```python
syslog.openlog(facility=syslog.LOG_AUTH)
syslog.syslog(syslog.LOG_INFO, "[HOWDY] Attempting facial authentication for user " + pamh.get_user())
...
syslog.syslog(syslog.LOG_NOTICE, "[HOWDY] Failure, timeout reached")
...
syslog.closelog()
```

Bubsbilby:
I'm having exactly the same issue. It's also preventing me from unlocking the lock-screen with howdy, but the password still works. The login screen still works as expected however. I've used both fixes found in the common issues page.

ramiel:
I can confirm the same error

boltgolt:
It seems that using `LOG_AUTH` also sets the priority to `LOG_EMERG`, even though that's not [documented](https://docs.python.org/3/library/syslog.html) and does not happen on my system. However, using `LOG_NOTICE` causes the log messages to end up in the general system log, not the authentication log

[...]

PR #406, opened 2020-07-29
Improve use of syslog

ponyisi:
This should fix issue #385 .

The existing code was effectively always calling syslog with EMERG priority; on some distributions (e.g. OpenSUSE) this would result in messages on every console and the GUI notification area. I have reduced the severity of the messages according to my best guess of what they should be; the actual login success messages are now at INFO priority.


boltgolt:
While your changes are good, we settled on using `syslog.LOG_NOTICE | syslog.LOG_AUTH` so the lines would still go to the auth log (see #374). This is now implemented on the dev branch

However, you also added some other good stuff like opening and closing the log properly, would it be possible for you to apply those changes to the new dev branch?

ponyisi:
Hi - actually this PR would also have sent messages to the auth log - if the `openlog()` call is made with `LOG_AUTH`, it's not necessary to include this info for every `syslog()` call. I'm happy to update but I'd prefer to do the setting of `syslog.LOG_AUTH` once in `openlog()` and keep only `syslog.LOG_NOTICE` in the syslog() calls, if that's ok with you?

boltgolt:
Thanks a lot!

PR #420, opened 2020-08-14
Fix negative sleep time error

eteq:
It's not clear to me in what circumstances this occurs, but on my machine at least in certain circumstances the test script errors with a traceback that indicates a negative sleep time was given.  This PR should fix that by just skipping the `sleep` call if this happens.

### DIFF
--- a/src/cli/test.py
+++ b/src/cli/test.py
@@ -182,7 +182,7 @@ try:
 		frame_time = time.time() - frame_tm
 
 		# Delay the frame if slowmode is on
-		if slow_mode:
+		if slow_mode and (frame_time > .5):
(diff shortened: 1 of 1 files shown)


boltgolt:
Thanks for the patch, but unfortunately this issue has also been fixed before in #401 which is on the dev branch. I'm sorry i haven't released that build yet, but many thanks for getting involved in the project!

eteq:
Oh, hah, in fact I was using the dev branch, but I didn't get around to submitting this PR for a few weeks after I installed it.  So really my bad for not submitting this when I first encountered it!  Thanks, @boltgolt

Issue #387, opened 2020-07-06
Privilege detection is broken when using PolicyKit


aleixpol:
I've been looking into implementing a Plasma front-end for Howdy (https://invent.kde.org/apol/plasma-howdy) although it's not working. For escalating privileges like we need for howdy (which is something we probably should be revised sometime) we generally use PolicyKit, but when we do so howdy complains about not being run under root for some reason.

An easy way to test it is this:
```
$ pkexec python /lib/security/howdy/cli.py list
Please run this command as root:

        sudo howdy list
```
----

Linux distribution (if applicable): ArchLinux
Howdy version: 2.6.0

GarvTambi:
Sometimes face detect successfully and sometimes it shows Unknown Error:1 Why?
Please help!

![dfg](https://user-images.githubusercontent.com/42083679/87365187-3d728080-c593-11ea-9776-8207838b0993.jpeg)

aleixpol:
Please open a separate issue for this, it's unrelated.

boltgolt:
Please stop spamming this in every issue, or i will have to remove you @GarvTambi

boltgolt:
A frontend for Howdy would be amazing! 

The issue is the way Howdy checks if it has been run with sufficient permissions. Right now it's a [simple check](https://github.com/boltgolt/howdy/blob/master/src/cli.py#L83) to see if the environment variable "SUDO_USER" has been set. Something more robust might be needed there.

Pointedstick:
You could maybe check to see if the UID is 0. That works for pkexec and also catches sudo too.

```
$ id -u
1000

$ pkexec id -u
0

$ sudo id -u
0
```

boltgolt:
Probably, fix is in and the GUI is done too

PR #452, opened 2020-10-12
Change detecting sudo priviliges to check if process euid is 0

espidev:
This should address #387


boltgolt:
Thanks! I'll clean up the line a little bit but this change seems to work perfectly

PR #455, opened 2020-10-22
Stop verification and print warning if video device does not exist.

bob8677:
I have a laptop with a camera kill switch. When trying to verify with Howdy with the camera disabled, I get the following python error.
`Camera path is not configured correctly, please edit the 'device_path' config value.
Exception ignored in: <function VideoCapture.__del__ at 0x7fab7dca6ee0>
  File "/usr/lib/security/howdy/recorders/video_capture.py", line 55, in __del__
    self.internal.release()
AttributeError: 'VideoCapture' object has no attribute 'internal'
Unknown error: 1`

As I sometimes disable the camera, this gets annoying. This change makes only the useful warning `Camera path is not configured correctly, please edit the 'device_path' config value` show up, and gives an option, `warn_no_device` to silence the warning.

### DIFF
--- a/src/config.ini
+++ b/src/config.ini
@@ -40,6 +40,9 @@ timeout = 4
 # Should be set automatically by an installer if your distro has one
 device_path = none
 
+# Print a warning if the the video device is not found
+warn_no_device = true
+
 # Scale down the video feed to this maximum height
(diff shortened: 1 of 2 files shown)


boltgolt:
Very useful change, thank you!

PR #463, opened 2020-11-03
Add feature for test to also do recognition

eteq:
To try to troubleshoot some lack-of-recognition problems I was having, I updated the test cli script to add a stage where it does face model comparison.

Two things I wasn't sure of in this implementation:
1. Because  `compare.py` isn't in a "real" python package, there's no straightforward way I could work out to use it directly in the "test" script.  So I had to copy-and-paste, which is problematic if the `compare.py` changes down the line.
2. There are now several parameters to the test command, but I didn't see an obvious way to pass thos in from `cli.py` with the way this is architected.  So I did environment variables.  Can change that if there's a desired idiom other than that, though.

### DIFF
--- a/src/cli/test.py
+++ b/src/cli/test.py
@@ -2,19 +2,28 @@
 
 # Import required modules
 import configparser
+import builtins
 import os
 import sys
 import time
 import dlib
 import cv2
+import json
 from recorders.video_capture import VideoCapture
+import numpy as np
 
 # Get the absolute path to the current file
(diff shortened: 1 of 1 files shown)


boltgolt:
Thank you so much! I like the changes a lot, it makes troubleshooting with the test command a lot more effective. I'm not 100% sure about the prefixed env variables though, i don't think a lot of people are going to use those and they almost feel like a "hidden" feature. Do you think all these variables are important to keep?

boltgolt:
Superseded by #504

PR #464, opened 2020-11-05
Fixed missing sphinx package for Arch

rafalgorczewski:
I suppose, that the second directory containing pam-python could be safely removed, since it can be found on official repos, but I'm not 100% sure.


rafalgorczewski:
It seems that it doesn't build pam_python.so

PR #465, opened 2020-11-05
python2-sphinx replaced with python-sphinx

rafalgorczewski:
Since python2-sphinx is no longer available, I replaced it with python-sphinx package, which seems to solve building issues.


boltgolt:
Makes sense, thanks!

PR #470, opened 2020-11-11
Changed final print statement to reflect changes

arifer612:
Arguments for the `disable` command are either `{0, false}`, or `{1, true}`. However, only if the argument is `1` will the print statement print "Howdy has been disabled". It should, by right, also show if the argument was `true`. To fix it, I noticed that the 2 possible sets of arguments fall into either of 2 groups characterised by the variable `out_value`. I thus suggest that the variable be used in the conditional for the final print statement.


boltgolt:
Thank you!

PR #472, opened 2020-11-14
Update arch packages to work with python-sphinx instead of python2-sphinx

Giannie:


boltgolt:
Thanks for making an PR! I just merged #464 before i saw this one, do you think merging your extra changes are necessary too?

Giannie:
There are some changes that should be added. Currently the license file is not installed and it should be. This is included in commit https://github.com/boltgolt/howdy/pull/472/commits/b3c0ac291b572fe53950e6b4ee68754f9016049a.

The python-pam library is also missing a dependency on pam. This is unlikely to cause an issue as I'm sure most installations will include pam, but it should also be included, see commit https://github.com/boltgolt/howdy/pull/472/commits/d253522c1b3fd7221058b2030f56eeb042e61053.

Neither of these are likely to cause installation issues, but should be resolved anyway.

As I said before, if you'd like someone to take on maintenance of the AUR packages, I'd be happy to.

boltgolt:
@Giannie thanks so much for your contributions! I'm going to merge them in through #477 as it does not have any conflicts

PR #477, opened 2020-12-03
Bump AUR to 2.6.1

kageurufu:
This includes adding the LICENSE to pam-python, thanks to @Giannie for that

This does supersede #472, and matches what is current on the AUR


boltgolt:
Thanks! Wouldn't it be a good idea to also add pam as a dependency (d253522) as suggested in #472?

kageurufu:
I was assuming `pam` was part of the `base` metapackage, but it looks like its a transitive dependency. I'll add it to pam-python

boltgolt:
Thanks a lot!

Issue #478, opened 2020-12-07
Add option for user authentication when using sudo


DonFlymoor:
I've had to disable howdy being able to authenticate as root, because I was afraid that some malicious application would gain root privileges that way.
What happens:
```bash
user@computer:~$ sudo su
... howdy athenticates ...
root@computer:/home/user#
```
The webcam turns on, and when face verification succeeds, it logs on as root. This has serious security implications, as malicious program could easily gain root privileges, possibly without the user noticing.

This is what I would like to see on the command line:
```bash
user@computer:~$ sudo echo "testing"
Do you wish to allow echo to be run as root [y/n]:y
... howdy athenticates ...
testing
```

And when the command is not run in the terminal a popup should appear, although I think that may already happen.

Issue #456, opened 2020-10-24
Skip facial detection after password entered


ralphocdol:
There are times when my face isn't detected properly due to poor lighting environment.

In `pkexec` case or the GUI authentication, the facial recognition waits for the face to be detected or timeout to reach even after the password was already entered.

In `sudo` in terminal/console, the facial detection is asked first before the password (when it fails).

It would be better if the password is asked while the facial recognition is detecting the face then:
- Skip the face detection if password is entered
- Skip password when face is detected

I don't know if this can be achieved but I think this could help us do things seamlessly.

Thanks :)

FedeCuci:
+ 1 for me. Came here to open the exact same thing.

X-Ryl669:
As far as I've understood, it's not possible. This is not a limitation in Howdy but in PAM as it's not multithreaded, so it tries one module after the other in order. There used to be hack by allowing, in one module, multiple input (like stdin (keyboard) input + other method), but it breaks the security of PAM where each module is audited for what it does.

So, you should report a bug to PAM developers to allow multi-thread/multi-process authentication. The same issue arise with fprint for fingerprint authentication.

FedeCuci:
Is this their github repo: https://github.com/linux-pam/linux-pam? @X-Ryl669

X-Ryl669:
Yes, but for an example implementation of a "parallel" implementation, see [here](https://unix.stackexchange.com/questions/207360/use-multiple-pam-login-methods-at-once). It requires hacking the pam module to implement stdin/password capture and, honestly, it looks like it's a hack than a genuine operation.

FedeCuci:
I'd be happy to open an issue on their GitHub, I just wouldn't really know how to explain myself as I don't know how the PAM system works apart from you said above @X-Ryl669.

X-Ryl669:
I've started an issue on their repo. Cheers!

FedeCuci:
That's nice of you, thanks! @X-Ryl669

Issue #99, opened 2018-11-19
Any plans to speed up recognition?


schrodog:
Right now when I am using howdy with certainty = 5, it usually cost more than 2-3 seconds to authenticate, which is considerably slower than Windows hello (average < 2s speed). Any plan to further improve algorithm?

dmig:
For me noticeable speedup was gained by rebuilding `dlib` with full optimisations (see it's manual). Using CUDA also should give performance boost.

dmig:
Oh, btw, using normal cam (not IR one) slows down recognition, but gives more reliable result

dmig:
> ... rebuilding `dlib` with full optimizations...

Looking at `debian/postinst`, how it builds `dlib`, I realized, that I got deeper: I installed one of `dlib` optional dependencies, OpenBLAS, which is pretty well optimized and gained me better recognition speed.
But there are more options:
* [Intel® MKL](https://software.intel.com/en-us/mkl) (proprietary, might outperform OpenBLAS)
* [OpenBLAS](https://github.com/xianyi/OpenBLAS) itself, takes advantage of newer CPU instructions
* [BLAS from lapack](http://www.netlib.org/lapack/)
* [clBLAS](https://github.com/clMathLibraries/clBLAS) OpenCL backed
* [cuBLAS](https://developer.nvidia.com/cublas) CUDA backed

All of them except first are available as packages in standard repositories. By default, not finding any, `dlib` will issue a warning at configure stage and allow you to proceed, probably using some fallback code.

I'll make some tests and rewrite installation script later to choose best available libraries. I have a good feeling.

boltgolt:
Nice work! Does it speed up the per-frame recognition times or the face_recognition import? How much does installing OpenBLAS contribute to installation duration?

dmig:
I can't tell how much, since I didn't measure. I'll do that later. Libraries are installed as packages from Ubuntu repos.

boltgolt:
I've also been thinking about offloading the actual recognition to an (optional) deamon. However, compare.py uses about 217MB of memory right now which i think is a bit much to keep reserved. If we could do it though a cache it would be the best of both worlds though.

There's also the possibility of multithreading. Right now Howdy just uses 1 core and pushes it to maximum load, but doesn't utilize any of the other 7 on my machine because everything is running in a single thread. I think splitting the recognition into a separate process would also allow us to start multiple instances, increasing the amount of frames we can process.

dmig:
[...]

PR #484, opened 2020-12-26
Add native PAM module

musikid:
# Summary
This pull request introduces a native PAM module (written in C++) which waits for user input and asynchronously perform the face comparison.

Based on #9, I first tried to wait for the first event which responds, but since `pam_get_authtok` (the function which ask the user password) is blocking on the application-side, we have to wait for the user to fill the password and can't perform authentication on background.

What approach should we use then?

# Related
#99, #9, #478, #456


boltgolt:
Thank you so much for this implementation! I tried writing this exact thing before but I'm just not good enough in C++, as you've probably read in one of the issues you've linked 

The blocking issue is an annoying one. I tried to figure out how fingerprint-gui managed to get that to work. The fingerprint PAM module accepts both a password or your fingerprint and authenthecates based on whatever it got first. Maybe the answer is burried in its (pretty hard to read) source?

MusiKid:
I read the FingerprintGui source code, and the method that it uses to circumvent the blocking issue is dirty...
It spawns an helper process which sends `Enter`  when the user validates the fingerprint. I'm not sure of how reliable it is since the user could change the focused application and the input could not be sent then to the prompt. We would also have to write an helper process, which increases the difficulty.
Our best chance would be to wait for linux-pam/linux-pam#301 (hopefully) getting a proper implementation.

boltgolt:
Waiting for the PAM implementation is going to take a long while though. Not only does it have to be coded and merged, it's going to take years to be reliably available on all the distros

I really like the suggestion @X-Ryl669 posted in that thread though, could we write our own internal version of `pam_any`? Where waiting for the user password and running howdy are separate threads?

MusiKid:
You are right about the time that it would take to wait before an implementation 😅️.
For `pam_any`, I don't think it's possible because like I said before,  `pam_get_authtok` is blocking on **application-side**. The problem is not writing a module which supports waiting asynchronously, but rather the fact that PAM functions cannot be canceled safely.

boltgolt:
Ah sorry, i misunderstood what you meant by blocking on the application-side in your first post. Very annoying that PAM has no way to make this work without a hack.

Do you think it would be acceptable to implement the hacky workaround (sending a keypress from the Howdy python thread) as an experimental config option that is disabled by default?

MusiKid:
It's definitely possible, but it would require more work and also granting root access to the Python process because if we want to simulate a key press, we need to use `/dev/uinput`, which by default is only accessible to root.

boltgolt:
I've done some small testing with the python `pam.py` and i seem to be able to access `/dev/uinput`. As far as i know PAM modules run with root-like rights

MusiKid:
PAM modules runs with root rights **only** if the application is [setuid](https://en.wikipedia.org/wiki/Setuid) and the owner is set to root (see http://www.linux-pam.org/Linux-PAM-html/mwg-see-programming-sec.html). Even if it's uncommon and I am not aware of a real use case, it's possible to run a PAM module without root rights. You can test it easily with [pamtester](http://pamtester.sourceforge.net/) and `geteuid()`.

boltgolt:
Well, that solves the `/dev/uinput` problem then, right? The current implementation runs python as root too although that might not be ideal

MusiKid:
I think yes. The only problem now is how to communicate to PAM module that the Enter key was pressed by the comparator. I'm thinking about implementing it like FingerprintGui, which means using a pipe between Python and the module.

MusiKid:
We also need to think about how we should handle the failure of the face recognition.

boltgolt:
I've been thinking about this over the last few days. We could simply use the returned status code from the python process to see if enter was pressed or not. Certain codes should only be used if enter is pressed successfully. I've made a quick chart:

image

MusiKid:
Nice chart!
It should work now... I put some instructions in the README to build it. It needs way more testing to be considered stable but it is usable at least.

MusiKid:
Nice chart!
It should work now... I put some instructions in the README to build it. It needs way more testing to be considered stable but it is usable.

ccat3z:
Perhaps asking for the password in a new thread would avoid simulating enter press. Just cancel password prompt thread after howdy python process was returned.
I tried to use pthread to wrap `pam_get_authtok` to an asynchronous version in C and it seems work. Maybe useful for this pr.
See [pam_async_ext.h](https://github.com/ccat3z/pam_bio/blob/master/pam_async_ext.h) and [pam_async_ext.c](https://github.com/ccat3z/pam_bio/blob/master/pam_async_ext.c).

MusiKid:
I'm confused...
It is written nowhere that the conversation function (used by `pam_get_authtok` under hood) is a cancellation point and yet, doing what said @ccat3z works! (not all time though)
Do you encounter a problem with `sudo`? It needs the enter key press on my machine.

MusiKid:
Thanks @ccat3z for the tip BTW!

MusiKid:
It seems to work only on some programs (only pamtester AFAIK).

ccat3z:
> I'm confused...
> It is written nowhere that the conversation function (used by `pam_get_authtok` under hood) is a cancellation point and yet, doing what said @ccat3z works! (not all time though)
> Do you encounter a problem with `sudo`? It needs the enter key press on my machine.

No such problem on my laptop. sudo/pkexec/gdm works without press enter key.

MusiKid:
Nevermind, it was just a stupid bug. Thanks a lot for the tip!

boltgolt:
This is amazing input @ccat3z! I think we're branding this as experimental anyway so if it does not work in some instances that's okay

ccat3z:
After several days of daily use, I found that it is not always safe to cancel `conv`'s threads. As @MusiKid said in https://github.com/linux-pam/linux-pam/issues/301#issuecomment-762245349, `conv` is not always a cancellation point and resources may not be released too.
Here are some instances:
1. `pamtester` or other cli applications like `sudo` will turn off echo when retrieve password (`stty -echo`), input will not print after pam success. (`pamtester -v test-howdy username authenticate; read` and type in anything.)
2. `ligthdm` with `lightdm-gtk-greeter` hangup randomly after cancel `conv` thread. I'm not sure what happened, I'm not familiar with `lightdm`.

MusiKid:
After some testing, I also encountered some problems with `GDM`, probably because it tracks the wrong child process.
I need to do more debugging to check if it doesn't have other hidden bugs before finalizing the module.

boltgolt:
I'm not as familiar with C++ as you are, and i'm getting some really strange errors while trying to run `meson compile`. if i try to run the command as specified in the readme (`meson compile build`) then i get `ERROR: Can't invoke target 'build': target not found`.

Some of these errors are really strange, like `error: ‘function’ was not declared in this scope`. I feel like i'm compiling against the wrong version? What am i doing wrong here?

 <details>
  <summary>Click to open the complete command output</summary>

```
Found runner: ['.local/bin/ninja']
ninja: Entering directory `.'
[3/6] Compiling C++ object libpam_howdy.so.p/main.cc.o
FAILED: libpam_howdy.so.p/main.cc.o 
c++ -Ilibpam_howdy.so.p -I. -I.. -I../subprojects/inih/cpp -I/usr/include -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -g -fPIC -pthread -DBOOST_FILESYSTEM_DYN_LINK=1 -DBOOST_ALL_NO_LIB -MD -MQ libpam_howdy.so.p/main.cc.o -MF libpam_howdy.so.p/main.cc.o.d -o libpam_howdy.so.p/main.cc.o -c ../main.cc
In file included from /usr/include/c++/5/chrono:35:0,
                 from ../main.cc:16:
/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support \
  ^
../main.cc:79:3: warning: identifier ‘nullptr’ is a keyword in C++11 [-Wc++0x-compat]
   return conv(1, &msgp, &resp_, nullptr);
   ^
../main.cc:40:1: warning: scoped enums only available with -std=c++11 or -std=gnu++11
 enum class Type { Howdy, Pam };
 ^
../main.cc:42:29: error: ‘function’ has not been declared
 int on_howdy_auth(int code, function<int(int, const char *)> conv_function) {
                             ^
../main.cc:42:37: error: expected ‘,’ or ‘...’ before ‘<’ token
 int on_howdy_auth(int code, function<int(int, const char *)> conv_function) {
                                     ^
../main.cc: In function ‘int on_howdy_auth(int, int)’:
../main.cc:47:66: error: ‘conv_function’ was not declared in this scope
       conv_function(PAM_ERROR_MSG, "There is no face model known");
                                                                  ^
../main.cc:61:61: error: ‘to_string’ was not declared in this scope
                     string("Unknown error:" + to_string(code)).c_str());
                                                             ^
../main.cc: At global scope:
../main.cc:68:18: error: ‘function’ was not declared in this scope
 int send_message(function<int(int, const struct pam_message **,
                  ^
../main.cc:69:61: error: expression list treated as compound expression in functional cast [-fpermissive]
                               struct pam_response **, void *)>
                                                             ^
../main.cc:68:27: error: expected primary-expression before ‘int’
 int send_message(function<int(int, const struct pam_message **,
                           ^
../main.cc:71:18: error: expected primary-expression before ‘int’
                  int type, const char *message) {
                  ^
../main.cc:71:28: error: expected primary-expression before ‘const’
                  int type, const char *message) {
                            ^
../main.cc:71:47: error: expression list treated as compound expression in initializer [-fpermissive]
                  int type, const char *message) {
                                               ^
../main.cc:71:49: error: expected ‘,’ or ‘;’ before ‘{’ token
                  int type, const char *message) {
                                                 ^
[4/6] Compiling C++ object subprojects/inih/libINIReader.a.p/cpp_INIReader.cpp.o
ninja: build stopped: subcommand failed.
```

</details>

MusiKid:
I made a typo in the build command, it's supposed to be `meson compile -C build` and not `meson compile build`. Can you remove the `build` folder and retry?

boltgolt:
I've been working on this a bit more, it seems that i have to add `-std=c++11` to all cpp commands in build.ninja. I guess i'm using an older version?

```
error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
```

MusiKid:
I added the flag in Meson, but it's strange. I never had any warning about this.

boltgolt:
Finally got it fully working on my system! `python` does not always mean python3, and on my old Ubuntu 16.04 system it still points to python2 unfortunately. I've seen many issues over the years with incorrect python commands, so i've just linked it to the `/usr/bin/python3` executable directly now.

Your implementation of the simultaneous password typing and howdy authentication works flawlessly on my machine so far!

theol0403:
Been a howdy user for a year now, and I have to say, this really improves the experience! Not only is it now incredibly responsive and quick, it also seems to make howdy work in many more places like polkit. Thanks for this!

One note is that on my machine (fedora 34), howdy is installed in `/lib64/security/howdy` - but in this PR, `/lib/security/howdy` is hardcoded. I had to softlink the directory to get it to work.

And finally, for other people wanting to try this, it took me a bit of experimentation to figure out that the proper command to place in the PAM config files is `auth sufficient /lib/security/pam_howdy.so`.

Thanks!

boltgolt:
Howdy _should_ have always been in `/lib64/security`, but due to limitations in the use of `python_pam` it had to be in `/lib`. I'm planning to move it to `/lib64` in release 3.0.0 which is going to be a breaking release anyway

src/pam/main.cc
```diff
@@ -155,10 +171,11 @@ int identify(pam_handle_t *pamh, int flags, int argc, const char **argv,
   posix_spawn_file_actions_init(&file_actions);
   posix_spawn_file_actions_addclose(&file_actions, STDOUT_FILENO);
   posix_spawn_file_actions_addclose(&file_actions, STDERR_FILENO);
-  const char *const args[] = {"python", "/lib/security/howdy/compare.py",
+  const char *const args[] = {"/usr/bin/python3", "/lib/security/howdy/compare.py",
                               user_ptr, nullptr};
   pid_t child_pid;
-  if (posix_spawnp(&child_pid, "python", &file_actions, nullptr,
+
+  if (posix_spawnp(&child_pid, "/usr/bin/python3", &file_actions, nullptr,
```
MusiKid: We don't need to make python3 absolute since `posix_spawnp` already search the file in `$PATH` for us.

boltgolt: That's what i thought, but i learned the hard way that `python3` changes in `$PATH` when in a pyenv, potentially breaking howdy. Hardcoding also fixes a vulnerability where an attacker could simply overwrite the python3 path to any executable that returns code 0 and bypass Howdy completely

MusiKid: I wasn't aware of it. I reverted my commit.

boltgolt:
Thanks for your continued work on this! I've been running this patch locally over the last few weeks and i've noticed some significant drawbacks. For example, there's now 2 identical password fields on my greeter and when my face is verified i still have to press enter in one of them. There's also the more breaking issue where gnome-screensaver enters an infinite crash-and-start loop if i enter an empty password.

I would like to propose a change to the way in which we implement this. I think we should introduce a config variable (`async_pam`?) with 3 possible values: 

- `"off"`, which makes Howdy behave like it did in 2.x.x, where the pam module just spawns the python process 
- `"workaround"`, which uses the strategy where howdy sends an enter event to get around the blocking issue
- `"native"`, which uses what the thread solution we currently have in there PR to get around the blocking issue

Looking at the current issues on my system, i don't think we can set the native strategy as default for now. I would prefer setting the workaround strategy as default, and failing silently. This way new users can have the benefit of a working async solution without the side effects i currently see. Would love to hear your thoughts on this.

MusiKid:
Sorry for the (long time) absence!
I'm OK with using a config variable to select which strategy to use. But I'm not sure setting the workaround as default is the way to go, since it input <kbd>Enter</kbd> automatically and so can fail silently if the user change the focused window before it is taken into account.

boltgolt:
@MusiKid no problem! It's all volunteer work of course 

I hear what you're saying there and a missed enter would definitely be annoying. The other side is that the "native" strategy crashes gnome-screensaver on Ubuntu 16.04 and i don't think i can ship it like that by default. It sucks but that's unfortunately how PAM was implemented there.

The worst failure mode for the "workaround" strategy is that the user has to press enter manually.
The worst failure mode for the "native" strategy is that the GUI enters a restart loop and the user has to drop into the TTY to restart the whole window manager.

As i have to rewrite the readme for 3.0.0 anyway, maybe i can add a section after installation explaining why changing to "native" might be good thing?

williamschey:
pam_python did not work for polkit for me. It seemed to have an issue with the subprocess. This seems to work perfectly, although I had to change the directory for the config and script. In Fedora, the files live in the lib64 directory. Nice work though!

Any chance this could get merged? Happy to help if you want.

boltgolt:
I'm also really eager to merge this (basically the only thing delaying 3.0.0 with an GTK UI etc) but I'm too busy at the moment and so is the original author. 

I believe the only thing that really needs to be done now is to add the configuration option discussed above. That should switch between what's currently implemented in this PR ("native" thread killing), what was implemented a few months ago ("workaround" sending an enter keypress), and the current Howdy behaviour where you can't enter a password at the same time. 

Feel free to try to build this if you have the time.

archerdm:
Hey everyone. I've been using howdy for a while and thought I'd try this version. Does anyone have this working with KDE? For me, it does not work with the KDE screen locker (kscreenlocker_greet) at all. In fact even my password gets denied when this is enabled.

The other thing I noticed is that when it's used with sudo, terminal input is not visible until I issue "stty echo". Although, it's probably not secure to use face unlock with sudo anyways.

Not sure if I'm smart enough to figure it out, but I'm happy to test. Unlocking is much faster with this new version so I'm excited about the future of this project.

boltgolt:
I have also experienced echo being turned off when using this patch, seems to happen because the PAM process asking for a password is interrupted and does not re-enable echo

MusiKid:
Hi!
Just to say that I'm still working on it. You can check the (very experimental) progress [here](https://github.com/MusiKid/howdy/tree/experimental).

boltgolt:
Good to hear! I'd love to get this change through before shipping 3.0.0

KairuiLiu:
it seems that the echo not be hided
version : Howdy 2.6.1
深度截图_选择区域_20210728231759

MusiKid:
I have to check the tests and the packages building, but most of the work is done!

boltgolt:
Nice! Some simple testing i did locally went really smooth, at least with the "input" option

src/pam/main.cc
```diff
@@ -0,0 +1,394 @@
...
+                   (char *const *)args, nullptr) < 0) {
+    syslog(LOG_ERR, "Can't spawn the howdy process: %s", strerror(errno));
+    return PAM_SYSTEM_ERR;
+  }
+
+  mutex m;
+  condition_variable cv;
+  Type confirmation_type;
+  // TODO: Find a clean way to do this
```
iamkroot, line 244: Seems like a good usecase for [`std::atomic`](https://en.cppreference.com/w/cpp/atomic/atomic) [`Compare and Set`](https://en.wikipedia.org/wiki/Compare-and-swap).

musikid: Thanks for the suggestion!

src/pam/main.cc
```diff
@@ -0,0 +1,394 @@
...
+  if ((pam_res = pam_get_user(pamh, (const char **)&user_ptr, nullptr)) !=
+      PAM_SUCCESS) {
+    syslog(LOG_ERR, "Failed to get username");
+    return pam_res;
+  }
+
+  posix_spawn_file_actions_t file_actions;
+  posix_spawn_file_actions_init(&file_actions);
+  // We close standard descriptors for the child
+  posix_spawn_file_actions_addclose(&file_actions, STDOUT_FILENO);
```
iamkroot, line 226: Is there a particular reason for closing stdout/stderr? This is actually causing a crash when we have `end_report = true` set in the config and the python script tries to print the diagnostics. It crashes because stdout is not writable, meaning that the pam module assumes that the auth failed.

src/pam/main.cc
```diff
@@ -0,0 +1,394 @@
...
+      return on_howdy_auth(howdy_status, conv_function);
+    }
+  }
+
+  // The branch with Howdy confirmation type returns early, so we don't need an
+  // else statement
+
+  // Again, we need to be sure that we're not going to block forever if the
+  // child has a problem
+  if (child_task.wait(1.5s) == future_status::timeout) {
```
iamkroot, line 326: If we have already received user input in textual form, why is there a need to keep waiting for facial recog? Shouldn't it be killed immediately?

src/pam/main.cc
```diff
@@ -0,0 +1,394 @@
...
+  {
+    unique_lock<mutex> lk(m);
+    cv.wait(lk);
+    final_type = confirmation_type;
+  }
+
+  if (final_type == Type::Howdy) {
+    // We need to be sure that we're not going to block forever if the
+    // child has a problem
+    if (child_task.wait(3s) == future_status::timeout) {
```
iamkroot, line 289: Will we ever have to wait so long here? `final_type == Howdy` will only be satisfied when the python process has exited, and all that remains is for the task to return the status. Considering that `future::get` is a blocking operation, you may simply want to call that instead of trying to stop the task.

src/pam/main.cc
```diff
@@ -0,0 +1,394 @@
...
+    pass_task.stop(false);
+  }
+
+  char *token = nullptr;
+  tie(pam_res, token) = pass_task.get();
+
+  if (pam_res != PAM_SUCCESS)
+    return pam_res;
+
+  int howdy_status = child_task.get();
```
iamkroot, line 343: What will this value be when the child task is killed due to timeout? Seems like undefined behaviour.

src/pam/main.cc
```diff
@@ -0,0 +1,394 @@
...
+    final_type = confirmation_type;
+  }
+
+  if (final_type == Type::Howdy) {
+    // We need to be sure that we're not going to block forever if the
+    // child has a problem
+    if (child_task.wait(3s) == future_status::timeout) {
+      kill(child_pid, SIGTERM);
+    }
+    child_task.stop(false);
```
iamkroot, line 292: Considering that `future::get` is a blocking operation, you may simply want to call that instead of trying to stop the task.

src/pam/main.cc
```diff
@@ -0,0 +1,394 @@
...
+                 int type, const char *message) {
+  // No need to free this, it's allocated on the stack
+  const struct pam_message msg = {.msg_style = type, .msg = message};
+  const struct pam_message *msgp = &msg;
+
+  struct pam_response res_ = {};
+  struct pam_response *resp_ = &res_;
+
+  // Call the conversation function with the constructed arguments
+  return conv(1, &msgp, &resp_, nullptr);
```
iamkroot, line 115: > Hey everyone. I've been using howdy for a while and thought I'd try this version. Does anyone have this working with KDE? For me, it does not work with the KDE screen locker (kscreenlocker_greet) at all. In fact even my password gets denied when this is enabled.
> 
> The other thing I noticed is that when it's used with sudo, terminal input is not visible until I issue "stty echo". Although, it's probably not secure to use face unlock with sudo anyways.
> 
> Not sure if I'm smart enough to figure it out, but I'm happy to test. Unlocking is much faster with this new version so I'm excited about the future of this project.

Figured this out. `kcheckpass`, which is responsible for doing the auth with PAM on behalf of the KDE screenlocker, is getting a segfault in the conversation function: https://github.com/KDE/kscreenlocker/blob/084e6ab3b700b0354ec8d85d2e690773b51782b4/kcheckpass/checkpass_pam.c#L53.

The reason is that the pam module is supposed to send a user-defined `appdata_ptr` as the last argument to the `conv` function, but in our case, it is set to `nullptr` here.

All that needs to be done is to extract this `appdata_ptr` from the pam `conv` struct and pass it to this callback. This solves the crash.

iamkroot:
However, the KDE screenlocker still isn't working as expected. It only starts the auth _after_ Enter is pressed. But as soon as that happens, our pam `child_task` returns with an empty `auth_tok_ptr` and we kill the python subprocess since we think the user entered a password via text input.

I'm not sure what the correct solution is.
1. Should we check for "empty" `auth_tok_ptr`, and not kill the python process if that is the case? This would fix the KDE case, but not sure of any potential side-effects. 
2. Should KDE screenlocker be updated to start the PAM auth immediately after locking? This is kinda similar to Windows behaviour. And I would prefer to not have to press the Enter key every single time I want to unlock my screen. But the drawback is having to do this for potentially every DM's screenlocker, which would require extensive code changes.

boltgolt:
Thanks for digging into the code @iamkroot! I would say that we can keep python running on empty strings. If a user has such an extremely easy way of logging in, why would they want to slow it down by using Howdy? There are probably some people with edge cases somewhere but in the future we could make it a config option if needed

As for the PAM behavior of some greeters and lockscreens: I have given up getting them all to work perfectly as they show some very strange things like 2 seperate input boxes.

src/pam/main.cc
musikid: IIRC, it was indeed because the script was crashing when stdin was kept open. For stout and stderr, I don't remember a particular reason to close them, maybe just to take (too much) precautions.

src/pam/main.cc
musikid: Yes that was the reason. We should probably change the way to stop the python process for a cleaner one.

src/pam/main.cc
musikid: IIRC, it will be the signal which terminated it, something like SIGKILL, so definitely not a success.

src/pam/main.cc
musikid: Thanks for the digging! I will patch it ASAP.

src/pam/main.cc
musikid: You are definitely right, it's pointless to wait for or kill the child here. It must be an error that I made because of copy-pasting.

iamkroot:
I recommend running [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/) on the project (with default flags) to catch some common anti-patterns.

boltgolt:
Thanks for the updates @musikid! I've made a [debian package](https://launchpad.net/~boltgolt/+archive/ubuntu/howdy-beta/+packages) that should allow us to beta test this on a lot more devices soon, hopefully finally getting to the end and releasing 3.0.0

src/pam/main.cc
musikid: It's also because of the "input" workaround, when the compare process send an "Enter" input.

musikid:
> It might be good to run [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/) on the project (with default flags) to catch some common anti-patterns.

I added a configuration [file](https://github.com/boltgolt/howdy/pull/484/commits/e20a6ddd6847434ea6beeb9444ee4daa10cb54a7), but I'm not sure of what should be the flags.

musikid:
@boltgolt I saw that you added translations support, but I don't know where to find the files?
I need them to know if I can completely remove the Boost dependency.

src/pam/main.cc
```diff
@@ -0,0 +1,401 @@
...
+           errno);
+    return PAM_SYSTEM_ERR;
+  }
+
+  // NOTE: We should replace mutex and condition_variable by atomic wait, but
+  // it's too recent (C++20)
+  std::mutex m;
+  std::condition_variable cv;
+  std::atomic<ConfirmationType> confirmation_type(ConfirmationType::Unset);
+
```
musikid, line 272: Is it a problem if we force building against a recent compiler version? GCC and Clang (11) already have support for atomic waiting, but I'm not sure of how these versions are available.

boltgolt:
@musikid they need to be downloaded from Crowdin, i still need to do that and test with all languages. Here are the core files for Dutch and German:

[core (DE).zip](https://github.com/boltgolt/howdy/files/7918473/core.DE.zip)
[core (NL).zip](https://github.com/boltgolt/howdy/files/7918474/core.NL.zip)

musikid:
@boltgolt Thanks! I'll check that.

musikid:
Now the Enter press is handled directly in the module, rather than in the compare process. The module will try to send it in a loop, while the password prompt is active and until a fixed limit. If it didn't succeed, it will wait for the Enter key to be manually pressed or for an error to occur.

musikid:
The module can be merged now (finally!). I will open another PR to add the tests (because they are executed in a Docker container, which relies on the Debian package).
I'm not sure on how we should handle the translations, but you can generate the pot file with `meson compile pam_howdy-pot`.

src/pam/main.cc
```diff
@@ -0,0 +1,422 @@
...
+  // We want to stop the password prompt, either by canceling the thread when
+  // workaround is set to "native", or by emulating "Enter" input with
+  // "input"
+
+  // UNSAFE: We cancel the thread using pthread, pam_get_authtok seems to be
+  // a cancellation point
+  if (workaround == Workaround::Native && pass_task.is_active()) {
+    pass_task.stop(true);
+  } else if (workaround == Workaround::Input) {
+    if (geteuid() != 0) {
```
iamkroot, line 362: A nitpick: The fake device _can_ be created by any user which has access to `/dev/uinput`. So forcing it to be root seems weird. Why not just stat the file to see if the current proc has access?

src/pam/main.cc
```diff
@@ -0,0 +1,422 @@
...
+                      "press, waiting for user to press it instead"));
+    } else {
+      try {
+        EnterDevice enter_device;
+        for (int retries = 0;
+             retries < MAX_RETRIES &&
+             pass_task.wait(DEFAULT_TIMEOUT) == std::future_status::timeout;
+             retries++) {
+          enter_device.send_enter_press();
+        }
```
iamkroot, line 375: Maybe add a log message for when max retries are exceeded?

musikid: Thanks, I added your suggestion!

src/pam/main.cc
musikid: That's because I didn't know about `euidaccess` which can test access for the effective UID (instead of the real one). It's fixed now.

musikid:
Should I change the base branch for the `beta` one?

boltgolt:
Yes that would be perfect! The beta branch will be 3.0.0 in a future that's hopefully soon (next few months?)

Thank you so much for all the hours you have put into this PR and the repo as a whole!

musikid:
I was able to resolve the conflicts and rebase against `beta`, however it implies that I need to force-push the entire branch.
All metadata should be preserved, of course, but is it a problem if I do it?

boltgolt:
@musikid if that's the only solution then try it. This is force pushing your own branch right, not boltgolt/howdy?

musikid:
Yes, it will not touch it, of course! It's just that it will “pollute” a little more the log.

boltgolt:
Took 1,5 years but this is finally merged! Thanks so much to @musikid and all others in this thread

wmmc88:
This removes the dependency on python2(as mentionned in #535 ), right? Is there going to be a 3.0.0 release soon?

boltgolt:
I would like that too, but there is still work to be done on getting this out as a beta and after that we can do a general release, probably late summer?

boltgolt:
@muskid testing the beta release of 3.0.0 i'm getting this error in the PAM log:

```
pkexec: PAM unable to dlopen(/lib/security/howdy/pam_howdy.so): /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /lib/security/howdy/pam_howdy.so)
```

Seems to be one symbol that's the issue:

```
$ readelf --dyn-syms pam_howdy.so | grep '@GLIBCXX_3.4.29'
24: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND _ZNSt28__atomic_futex_uns@GLIBCXX_3.4.29 (11)
```

You're more comfortable with the whole c++ part of it, how would i go about using no version higher than GLIBCXX_3.4.28 that's available on Ubuntu 20.04?

musikid:
This shouldn't cause any problem AFAIK.

boltgolt:
Welll it's a fatal error, so howdy does not start up at all because of that

musikid:
You also need to rebuild it against the lower version of libstdc++. If I understand correctly, if you target Ubuntu 20.04, then you need to build the module on it too.

saschaeggi:
@musikid can we get a new release which include this new awesome feature?

musikid:
The release is still in the works but should come out ASAP.

Doomsdayrs:
> The release is still in the works but should come out ASAP.

do you need any help?

PR #489, opened 2021-01-02
Remove unneeded dependencies from Arch Linux

Jellyfrog:


boltgolt:
Thanks! `face_recognition` hasn't been used in a looong time

Issue #501, opened 2021-01-20
Archlinux package for pam-python still depends on python2


Bubbelb:
As I installed howdy on my Archlinux machine through AUR, I noticed a python2 dependency. Since python2 is deprecated, I just tried to change he dependency in the PKGBUILD to python. All went ok upon installation and in running. No problems whatsoever.

So I'd suggest to change the depencency from python2 to python.

boltgolt:
Probably a leftover from when pam-python only ran python2, thanks for the report

Bubbelb:
Oops, was too quick with my conclusion. I had python2 still installed when installing pam-python. So of course installation went ok.
Installation falis as python2 is not installed. At first I changed test.py and setup.py, so that they are running under python(3), not under python2. That doesn't help. The error now is::
````
==> Making package: pam-python 1.0.8-2 (do 21 jan 2021 17:13:01 CET)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> WARNING: Using existing $srcdir/ tree
==> Removing existing $pkgdir/ directory...
==> Starting build()...
make --directory doc
make[1]: Entering directory '/home/bbl/Downloads/AUR/src/pam-python-1.0.8/doc'
sphinx-build -b html -E . html
Running Sphinx v3.4.3
loading intersphinx inventory from http://docs.python.org/2.7/objects.inv...
intersphinx inventory has moved: http://docs.python.org/2.7/objects.inv -> https://docs.python.org/2.7/objects.inv
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
reading sources... [100%] pam_python                                                                                                                                                           
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] pam_python                                                                                                                                                            
generating indices... genindex done
writing additional pages... search done
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in html.
rm -f html/index.html && ln -s pam_python.html html/index.html
make[1]: Leaving directory '/home/bbl/Downloads/AUR/src/pam-python-1.0.8/doc'
make --directory src
make[1]: Entering directory '/home/bbl/Downloads/AUR/src/pam-python-1.0.8/src'
[...]

PR #502, opened 2021-01-20
Update PKGBUILD

Bubbelb:
Python2 is not needed anymore. Python (e.g. Python3 will suffice)
See issue #501


Bubbelb:
Does not work. More work needs to be done to update to python3.

PR #504, opened 2021-01-31
Model on test

### LINKED ISSUE #443: howdy test to show a face match
Is it possible to modify howdy test so it shows a red circle with no match and a green circle if it recognizes a face.

That way it will test howdy, not only if opencv recognizes faces

albfan:
This helps to understand what label on model howdy is detecting (at in what percentage)

I use face detection with several people on my family (to allow all of them to unblock)

When there's no model added, face is still detected. Not sure how to show that as red circle, while setting a blue circle if model label is detected. Looks when models are added, one is always the correct match.

The intention for blue, red circle (colors may vary) is to show when your face doesn't match anymore (different skin tone, glasses, hair style, beard) but still not working

To test I do this:

- remove models:

        sudo mv /usr/lib64/security/howdy/models/alberto{,.old}.dat

- restore models:

        sudo mv /usr/lib64/security/howdy/models/alberto{.old,}.dat
 
Fixes #443


albfan:
When a label is added to model would be interesting to take a snapshot to show what label matches with current face (to understand if face detection needs a new photo (like if you no longer have beards or viceversa)

src/cli/test.py
```diff
@@ -59,6 +62,23 @@ def print_text(line_number, text):
...
+try:
+	user = builtins.howdy_user
+	models = json.load(open(path + "/../models/" + user + ".dat"))
+
+	for model in models:
+		encodings += model["data"]
+except FileNotFoundError:
+	print("No face model known for the user " + user + ", please run:")
+	print("\n\tsudo howdy -U " + user + " add\n")
+	#sys.exit(1)
```
boltgolt, line 80: Did you want to remove this line initially? I think the current code is the best way of handling this error, notifying the user but not crashing

src/cli/test.py
```diff
@@ -80,7 +100,7 @@ def print_text(line_number, text):
 
 # Wrap everything in an keyboard interupt handler
 try:
-	while True:
+	while cv2.getWindowProperty("Howdy Test", cv2.WND_PROP_VISIBLE) > 0:
```
boltgolt, line 103: This looks really nice, will this properly close the window when you click the X?

boltgolt:
Hey, thanks for this great addition! I had a few questions which i've added as comments

src/cli/test.py
albfan: yes, as currently the code do not request you a model, if model is not present it will only mark a face but not resolve any tag on model

src/cli/test.py
albfan: Yes, I found that in other test for opencv and I find it really useful and intuitive

albfan:
Thanks! just let me know if there's anything else to rework.

For my use case (several faces from different people in model, show the model detected is interesting)

albfan:
@boltgolt anything else missing on this?

boltgolt:
Thank you for the changes and sorry it literally took years to merge them!

albfan:
Cool PR merged ever!

PR #506, opened 2021-02-10
add timeout_notice config option

nickrbclark:
Maintains current default behavior of printing the timeout message.
Makes it possible to run howdy "silently" from the user's perspective.


boltgolt:
Good addition, thanks a lot!

PR #528, opened 2021-04-06
Add python-opencv as a dependency for ArchLinux

jinnko:
In the AUR package details [GJRodenburg noted the following](https://aur.archlinux.org/packages/howdy/#comment-800292):

> Opencv has split of it's python parts since opencv 4.5.2-2, they are now available as python-opencv. This will need to be included as a dependency of Howdy.

Confirmed locally following an update that howdy was failing with:

```
  File "/lib/security/howdy/compare.py", line 18, in <module>
    import cv2
ModuleNotFoundError: No module named 'cv2'
Unknown error: 1
```

Following installation of [python-opencv](https://aur.archlinux.org/packages/python-opencv) howdy worked as expected.

### DIFF
--- a/archlinux/howdy/.SRCINFO
+++ b/archlinux/howdy/.SRCINFO
@@ -13,6 +13,7 @@ pkgbase = howdy
 	depends = python3
 	depends = python-dlib
 	depends = python-numpy
+	depends = python-opencv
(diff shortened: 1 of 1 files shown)


boltgolt:
Thanks!

PR #529, opened 2021-04-09
more friendly release

wyapx:


boltgolt:
Hey, could you add what error this fixes?

wyapx:
When the camera is not found, it will raise an error
This error does not affect the main function, but it does not look good

wyapx:
go to #532

PR #531, opened 2021-04-12
Use os.geteuid instead of env SUDO_USER to determine if running as root

### LINKED ISSUE #387: Privilege detection is broken when using PolicyKit

Zocker1999NET:
Fixes #387 

This will still detect the usage of sudo, but also of other tools like su, pkexec (or the rare case when users really log in and run applications as root …)

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -80,7 +80,7 @@ builtins.howdy_user = args.user
 
 # Check if we have rootish rights
 # This is this far down the file so running the command for help is always possible
-if os.getenv("SUDO_USER") is None:
+if os.geteuid() == 0:
 	print("Please run this command as root:\n")
 	print("\tsudo howdy " + " ".join(sys.argv[1:]))
 	sys.exit(1)


boltgolt:
Thanks for your work! Unfortunately #452 has already implemented this, and is on the dev branch. The master (main) branch is reserved for stable code so that's probably how you missed the previous PR. This change will be live with release 3.0.0 which is currently waiting on some breaking changes

Issue #458, opened 2020-10-28
ModuleNotFoundError: No module named 'ConfigParser'


HughPH:
Python never fails to give me a headache.

Howdy just wasn't even starting the webcam, so I ran `python3 /lib/security/howdy/pam.py` and I got `ModuleNotFoundError: No module named 'ConfigParser'`

So I installed it:
```bash
$ sudo pip3 install ConfigParser
Collecting ConfigParser
  Downloading configparser-5.0.1-py3-none-any.whl (22 kB)
Installing collected packages: ConfigParser
Successfully installed ConfigParser-5.0.1
WARNING: You are using pip version 20.2.1; however, version 20.2.4 is available.
You should consider upgrading via the '/usr/bin/python3 -m pip install --upgrade pip' command.
```

Then ran pam.py again:
```bash
$ sudo python3 /lib/security/howdy/pam.py 
Traceback (most recent call last):
  File "/lib/security/howdy/pam.py", line 10, in <module>
    import ConfigParser
ModuleNotFoundError: No module named 'ConfigParser'
```

One final thing to make sure it's in the right place...
```bash
$ sudo python3 -m pip install ConfigParser
Requirement already satisfied: ConfigParser in /usr/local/lib/python3.8/dist-packages (5.0.1)
```

And just a quick check of sys.path:
```bash
$ sudo python3 -c "import sys; print(sys.path)"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/local/lib/python3.8/dist-packages',  '/usr/local/lib/python3.8/dist-packages/dlib-19.16.0-py3.8-linux-x86_64.egg', '/usr/lib/python3/dist-packages']
```

Looks like the */usr/local/lib/python3.8/dist-packages* path is in there... so what's going on?

----

I've searched for similar issues already, and my issue has not been reported yet.

Linux distribution (if applicable): Ubuntu 20.04

Howdy version (`sudo howdy version`): 2.6.1

arifer612:
You need to have python 2 for it to work. This is because the `pam-python` module only operates on python 2. If you have python 2 you should not have to worry about `ConfigParser` not being installed because it should be part of the standard library.

boltgolt:
@arifer612 is exactly right! ConfigParser is not the issue here, did you notice any errors in `/var/log/auth.log`?

pavelkq:
[...]

Issue #445, opened 2020-09-15
INFRARED NOT WORKING 


magest2002:
respected sir   
                      in windows hello facial recognisation  my laptop has 2 ir blaster which help in -recognisation i cant figure out how to turn on  plz help me 

thanks

magest2002:
hpspectre@pop-os:~$ v4l2-ctl --list-devices --all
Synaptics RMI4 Touch Sensor (rmi4:rmi4-00.fn54):
	/dev/v4l-touch0

HP TrueVision FHD RGB-IR: HP Tr (usb-0000:00:14.0-5):
	/dev/video0
	/dev/video1

Driver Info:
	Driver name      : uvcvideo
	Card type        : HP TrueVision FHD RGB-IR: HP Tr
	Bus info         : usb-0000:00:14.0-5
	Driver version   : 5.4.44
	Capabilities     : 0x84a00001
		Video Capture
		Metadata Capture
		Streaming
		Extended Pix Format
		Device Capabilities
	Device Caps      : 0x04200001
		Video Capture
		Streaming
		Extended Pix Format
Priority: 2
Video input : 0 (Camera 1: ok)
Format Video Capture:
	Width/Height      : 640/480
	Pixel Format      : 'YUYV' (YUYV 4:2:2)
	Field             : None
	Bytes per Line    : 1280
	Size Image        : 614400
	Colorspace        : sRGB
	Transfer Function : Default (maps to sRGB)
	YCbCr/HSV Encoding: Default (maps to ITU-R 601)
	Quantization      : Default (maps to Limited Range)
	Flags             : 
Crop Capability Video Capture:
	Bounds      : Left 0, Top 0, Width 640, Height 480
	Default     : Left 0, Top 0, Width 640, Height 480
	Pixel Aspect: 1/1
Selection Video Capture: crop_default, Left 0, Top 0, Width 640, Height 480, Flags: 
Selection Video Capture: crop_bounds, Left 0, Top 0, Width 640, Height 480, Flags: 
Streaming Parameters Video Capture:
	Capabilities     : timeperframe
	Frames per second: 30.000 (30/1)
	Read buffers     : 0
                     brightness 0x00980900 (int)    : min=0 max=255 step=1 default=128 value=128
                       contrast 0x00980901 (int)    : min=0 max=255 step=1 default=32 value=32
                     saturation 0x00980902 (int)    : min=0 max=100 step=1 default=64 value=64
[...]

Issue #367, opened 2020-06-13
Dell XPS 9500 IR camera not recognized


ctrlbrk42:
Ubuntu 20.04 LTS
Dell XPS 15 9500 (2020 model)

Camera light turns on, indicating image recording -- but the IR sensors do not turn on.  Tried the config/reinstall a few times.

Not sure what command output I can include to diagnose, here is lsusb -v for what I think is the camera.

```
Bus 001 Device 003: ID 0c45:6d14 Microdia Integrated_Webcam_HD
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.01
  bDeviceClass          239 Miscellaneous Device
  bDeviceSubClass         2 
  bDeviceProtocol         1 Interface Association
  bMaxPacketSize0        64
  idVendor           0x0c45 Microdia
  idProduct          0x6d14 
  bcdDevice           92.67
  iManufacturer           2 Sonix Technology Co., Ltd.
  iProduct                1 Integrated_Webcam_HD
```

andrewmv:
I am successfully using Howdy on a 2020 Dell XPS 13 9300, though I'm using Ubuntu 18.04. I didn't need to change any of the default configuration in order for the camera to be recognized, though I did need to tweak the darkness settings for it to get successfully detect a face - my open PR #366 deals specifically with this issue.

Your problem description makes me suspect your RGB camera might have been detected as infrared. Try explicitly defining the camera device - you can use the following command with each of the devices in /dev/video* to try to determine which video device is correct:
`fswebcam -d /dev/videoX out.jpg`

boltgolt:
Agreed @andrewmv , have you tried this @ctrlbrk42 ?

FIUSHerson:
I actually have the same device running Pop, and the tweak I made to get this working was to enter `sudo howdy config`, and changing `device_path = /dev/video2`.

In case that specific device path doesn't work, you can try searching up your list of webcams and ir sensors with `ls -ltrh /dev/video*`

Hope this helps!

GarvTambi:
Sometimes face detect successfully and sometimes it shows Unknown Error:1 Why?
Please help!

![dfg](https://user-images.githubusercontent.com/42083679/87365792-c50cbf00-c594-11ea-98df-891aca31fe53.jpeg)

DasJott:
I also run a 2020 xps13 9300 on Manjaro Linux and it doesn't work out of the box.
In config the path for webcam is `none`.
[...]

Issue #312, opened 2020-02-09
Lenovo X1 - Carbon gen 7 - Ir Emitter issues


dockera:
Hi There,

Great work on howdy. Had an issue where the Chicony IR emitter was not functioning at all. After some research, compiled the IR toggle app in this repository. This ran effectively under sudo and I was able to use howdy during a session. However, the IR emitter was not toggled back on by default under Ubuntu 19.10. To get around this, I created a service inside systemd (/lib/systemd/system) called chicony.service. To create this, used the following steps

1. move into the appropriate directory - cd /lib/systemd/system
2. create the service file - sudo touch chicony.service
3. edit the service file - sudo nano ./chicony.service
4. edited the service file by inserting the content below

[Unit]
Description=chicony service

[Service]
ExecStart=/path to compiled ir-toggle-command/chicony-ir-toggle-master/chicony-ir-toggle on

[Install]
WantedBy=multi-user.target

6. Set permissions for the service to match the other services in the directory
7. ran -  sudo systemctl start chicony - this starts the service and runs the command
8. ran - sudo systemctl status chicony  - checks that the service ran successfully
9. ran - sudo systemctl enable chicony - sets the service to run at boot

This ensures that the IR emitter is toggled on, allowing howdy to run from boot

Hope someone finds this useful. Found that the emitter only ever flashes once, not sure how to replicate the pattern of usage under Windows whereby windows hello causes it to flash rapidly and repeatedly - this likley affects accuracy etc. 

Would be keen to see the IR toggle function set within howdy, would offer to help but not a real coder :-)

shivindera:
Just to add to the above answer. Information I feel could be useful:
- The repo for chicony-ir-toggle is - [chicony-ir-toggle](https://github.com/PetePriority/chicony-ir-toggle)
- If you follow the install steps, the executable file is located in `/usr/local/bin/chicony-ir-toggle`
- When you run `chicony-ir-toggle on`, the ir sensor feed goes through video2, not video0,
- Modify the file `/lib/security/howdy/config.ini` and change the line to `device_path = /dev/video2`
- Run `sudo howdy test` and you will see that the feed is coming from the ir sensor.

Note: I had to modify the dark_threshold to 100 in the config file for it to recognise my face using the IR sensor. @dockera did you have to do something similar?

LEXUGE:
[...]

PR #532, opened 2021-04-13
Resolve some problem

wyapx:
## Already fixed (tested on my computer)
#445 
## Not test
#458 
#367 
#312 
#269 
and I do something change


boltgolt:
Hey, thank you for your work but i can not accept these changes as they are right now. You are very welcome to fix stuff but this is a monster of a PR with over 1000 changed lines, including edits to a library file that should not be edited. 

I would suggest that you find what exactly you want to fix, change only what you need to change and open a new PR describing what you've changed and why. I hope you understand that i can't manage huge PRs like this, simply look at the merge conflict list.

boltgolt:
Also, Howdy does not officially support non-IR cameras. A config option like `is_ir_camera` therefore doesn't really fit.

wyapx:
OK, I will modify and resubmit

But I don’t recommend using tab as indentation, This does not comply with the PEP8 specification

PR #546, opened 2021-05-17
Update PKGBUILD to add python-opencv

xuanruiqi:
The previous PR #528 adds `python-opencv` as a dependency, but (erroneously) only in `.SRCINFO`. That is not right, it needs to be added to `PKGBUILD`.

BTW, `.SRCINFO` probably should not be checked in to the interface as it's meant to be generated from `PKGBUILD`. It is only used by the AUR interface to correctly display dependencies, etc.

### DIFF
--- a/archlinux/howdy/.SRCINFO
+++ b/archlinux/howdy/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = howdy
 	pkgdesc = Windows Hello for Linux
 	pkgver = 2.6.1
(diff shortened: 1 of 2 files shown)


xuanruiqi:
Forgot to bump pkgrel, which I just did.

boltgolt:
Implemented as commit ceca88a as the changes to beta branch were too large to merge. Sorry for the enormous delay on this!

Issue #263, opened 2019-10-13
Howdy CLI Not Starting on Ubuntu 19.04


RD1K:
I just installed Howdy using the commands provided in the README file here. It installed fine; if I enter `sudo apt install howdy` into the terminal it says `howdy is already the newest version (2.5.1)`. However, when I try to add a face with `sudo howdy add`, or even if I just enter in `howdy` or `sudo howdy config`, I get this output: `howdy: command not found`. Why might this be?

I've searched for similar issues already, and my issue has not been reported yet.

Linux distribution (if applicable): Ubuntu 19.04 Desktop

Howdy version: 2.5.1

boltgolt:
Weird, what does `whereis howdy` output for you?

RD1K:
It outputs "howdy: /use/share/man/man1/howdy.1.gz"

Bound3R:
I have the same problem, I really wanted to try this package. :cry:

atharvaunde59:
I'm also facing same issue on Ubuntu 19.10


root@atharva-Inspiron-5559:/usr/bin# sudo howdy add
sudo: howdy: command not found
root@atharva-Inspiron-5559:/usr/bin# whereis howdy
howdy: /usr/share/man/man1/howdy.1.gz

Bound3R:
Hello, I followed these steps to install it manually and this method worked for me you can see the thread of issue [#256 ](https://github.com/boltgolt/howdy/issues/265) 

on resumen i run this comands, maybe need sudo.
```
apt remove howdy
pip3 install --upgrade pip
sudo dpkg --install howdy_2.5.1.deb
```

RD1K:
@Bound3R Where did you get the .deb file from?

Bound3R:
> @Bound3R Where did you get the .deb file from?

Directly from the [releases page](https://github.com/boltgolt/howdy/releases), to be more specific use version [_2.5.1_ ](https://github.com/boltgolt/howdy/releases/download/v2.5.1/howdy_2.5.1.deb).

RD1K:
I switched distros to Manjaro a few days ago, so I will no longer be able to test any solutions. Therefore, I'm closing this thread.

matifali:
> apt remove howdy
> pip3 install --upgrade pip
> sudo dpkg --install howdy_2.5.1.deb

This worked for me, exact commands I used were
```
sudo apt remove howdy
sudo pip3 install --upgrade pip
[...]

PR #548, opened 2021-05-27
Updated installation instructions for Ubuntu

monees007:
Updated installation instructions for Ubuntu or Linux Mint.
Due to [Howdy CLI Not Starting on Ubuntu 19.04 #263](https://github.com/boltgolt/howdy/issues/263)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,15 +26,14 @@ Howdy is currently available and packaged for Debian/Ubuntu, Arch Linux, Fedora 
 **Note:** The build of dlib can hang on 100% for over a minute, give it time.
 
 ### Ubuntu or Linux Mint
-
+Download the .deb file from the [Releases page](https://github.com/boltgolt/howdy/releases) and open a terminal in that directory.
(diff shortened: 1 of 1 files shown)


boltgolt:
While i appreciate you trying to help others, i don't think making new users add a PPA but installing a 2 year old version of Howdy a deb is the best solution

boltgolt:
Additionally, a pip upgrade is already triggered during Howdy install

monees007:
The reason I suggested that change was that I was installing howdy from ppa
in Ubuntu 20.04, and it completed successfully. But in running `howdy` I
get, howdy is not installed. But when I run `sudo apt install howdy' it
shows the package is already on its latest version. I tried complete
removal and reinstallation, even that didn't work.
Then I completely removed it. And installed that .Deb with `dpkg`
It worked. After that I updated howdy with apt' ppa. And I got latest
version running smoothly.

Then I searched the issue and came across #548. And the `README.md`
described using Deb file in Debian system. Since Ubuntu is very much Debian
based. I did that changes.

Over to you.
Thanks and have a nice weekend.
Monees Chandra

PR #550, opened 2021-05-31
Show user's Full Name in Messages

### LINKED ISSUE #493: Add option to display Full Name in the "Identified face as ..." message.
Currently, Howdy displays a message with the user's id:

    "Identified face as user_id"

I wold like to request the ability to make Howdy look up the user's name, and display that instead. For example:

    "Identified face as Mickey Mouse"

I understand that multiple user id's on a system could have the same full name. Ideally, this feature should be configurable system-wide for Howdy, for those that want to use it on their personal machines.

robodair:
This is configurable using the core.display_full_name boolean key
Fixes #493

### DIFF
--- a/src/config.ini
+++ b/src/config.ini
@@ -27,6 +27,9 @@ disabled = false
 # computational power to run, and is meant to be executed on a GPU to attain reasonable speed.
 use_cnn = false
 
+# Display user Full Name in PAM Messages instead of Username
+display_full_name = true
+
 [video]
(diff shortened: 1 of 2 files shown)


boltgolt:
That's a lot of sider spam, disabled it completely for now.

Thanks for your patch! Unfortunately pam.py is being slowly rewritten into C++ over at #484. If you want your changes in version 3.0.0 they would have to be rewritten in C++ i'm afraid

PR #561, opened 2021-07-03
feat: support non-sudo like pkexec

kolbma:
With an additional uid check this works also with tools like `pkexec` to change to root permissions.


RealTheSunCat (approved):

boltgolt:
Thanks for your contribution! Unfortunately this has already been implemented on the dev branch

RealTheSunCat:
Whoops, didn't know that would appear there, sorry!
Love the work y'all are doing :)

PR #585, opened 2021-09-04
Retry if failed to open capture device

Dharmendra63:
Elementary os greeter https://github.com/elementary/greeter is opening
two parallel login requests when trying to login after lock. One of
these requests fail because the other one has captured the camera.
Hence this commit makes hodwy do retries when capturing the camera.

### DIFF
--- a/src/recorders/video_capture.py
+++ b/src/recorders/video_capture.py
@@ -6,6 +6,7 @@ import configparser
 import cv2
 import os
 import sys
+import time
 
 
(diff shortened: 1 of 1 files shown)


boltgolt:
Sorry for being so extremely late to this PR! While i really appreciate your patch, some users _want_ Howdy to fail when it it can't open the camera. For example, people that use a USB camera and only want the password prompt when it's not connected. I'm going to close this PR but will refer anyone here who has the same use case as you do

PR #590, opened 2021-09-18
style: removed pointless check, fixed typos

moonburnt:
- "if not" will already trigger on None, no need to specifically mention it
- Fixed typos presented in original code's comments


boltgolt:
Thanks for the small changes! 

However, I do not think the the very large change from spaces to tabs is needed. The project as always used tabs and I do not see a reason to switch to spaces now. PEP8 strongly recommends spaces (the given reason for that is that _most projects_ use spaces, which is important to some) but in no way is the code quality compromised by using tabs.

moonburnt:
> However, I do not think the the very large change from spaces to tabs is needed.

okay, now there should be only small changes

boltgolt:
Merged as 84ad5ddc52b9848a17e884c0630e0fae2cd83caf, original branch has since been deleted. Sorry for my slow pace and thanks for the patch!

PR #605, opened 2021-10-30
Fix some bug on GUI

EmixamPP:
I split the fixes into 3 commits, here the explenations :
1. https://github.com/boltgolt/howdy/commit/0e4aaa0c47ed0a90831fc071ced6de890128c29c For example, on fedora Howdy is installed in `lib64` instead of `lib`
2. https://github.com/boltgolt/howdy/commit/2e2c1ac1969e65c6ca992bb43e7233c2733d8d38. On many laptop the camera is split into many `/dev/video` descriptor. At least 2 for 'gray camera' and 'color camera'. But there may also be others that are "useless", it is impossible to read an image with these file descriptors.
3. https://github.com/boltgolt/howdy/commit/4f7c65615c4043df2245bd49162fef0e163f7bad Most of the time the camera list is not displayed, I swapped the display instruction order and since then it works in 100% of the cases.

PS: I'll try to start integrating linux-enable-ir-emitter with Howdy. If you're still up for it ?

### DIFF
--- a/howdy-gtk/src/onboarding.py
+++ b/howdy-gtk/src/onboarding.py
@@ -68,12 +68,22 @@ class OnboardingWindow(gtk.Window):
 		eventbox = self.builder.get_object("downloadeventbox")
 		eventbox.modify_bg(gtk.StateType.NORMAL, gdk.Color(red=0, green=0, blue=0))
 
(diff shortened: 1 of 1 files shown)


boltgolt:
Amazing, thank you for checking out the GUI! I'm still very much up for integrating `linux-enable-ir-emitter` cleanly, just a tad busy atm

PR #610, opened 2021-11-11
inverted error message

EmixamPP:
I unintentionally switched two error messages in the previous fixes.
Sorry !


boltgolt:
Thanks for the fix!

PR #611, opened 2021-11-11
linux-enable-ir-emitter integration

EmixamPP:
Here is the first implement of `linux-enable-ir-emitter` in `howdy-gtk`.

But I am not totally satisfied :
* `gnome-terminal` must be add as dependency
* because I run `linux-enable-ir-emitter` through `gnome-terminal`, it's impossible to get the exit code
* in resume: `howdy-gtk` just run `linux-enable-ir-emitter` without any feedback

The best would be implement a system which output in `howdy-gtk`, the output `linux-enable-ir-emitter`. And ask to press a button yes/no to respond to the input of `linux-enable-ir-emitter`. I have tried to implement this, but the solutions are really not simple and require much more work.

So, here is a first 'simple' implementation, you can see what it would look like. And maybe see for a better integration if this one doesn't fit to your ideas.

### DIFF
--- a/howdy-gtk/src/onboarding.glade
+++ b/howdy-gtk/src/onboarding.glade
@@ -452,7 +452,7 @@ but will take much longer to authenticate you</property>
                     <property name="visible">True</property>
                     <property name="can-focus">True</property>
                     <property name="receives-default">True</property>
-                    <signal name="clicked" handler="go_next_slide" swapped="no"/>
+                    <signal name="clicked" handler="leie_button_yes" swapped="no"/>
                   </object>
                   <packing>
                     <property name="expand">True</property>
@@ -466,7 +466,7 @@ but will take much longer to authenticate you</property>
                     <property name="visible">True</property>
                     <property name="can-focus">True</property>
                     <property name="receives-default">True</property>
-                    <signal name="clicked" handler="show_leie" swapped="no"/>
+                    <signal name="clicked" handler="leie_button_no" swapped="no"/>
                   </object>
                   <packing>
                     <property name="expand">True</property>
--- a/howdy-gtk/src/onboarding.glade
+++ b/howdy-gtk/src/onboarding.glade
@@ -1,56 +1,56 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.38.2 -->
 <interface>
   <requires lib="gtk+" version="3.12"/>
   <object class="GtkImage" id="iconcancel">
     <property name="visible">True</property>
-    <property name="can_focus">False</property>
-    <property name="margin_right">4</property>
+    <property name="can-focus">False</property>
+    <property name="margin-right">4</property>
     <property name="stock">gtk-cancel</property>
   </object>
   <object class="GtkImage" id="iconfinish">
     <property name="visible">True</property>
-    <property name="can_focus">False</property>
-    <property name="margin_right">5</property>
+    <property name="can-focus">False</property>
+    <property name="margin-right">5</property>
     <property name="stock">gtk-apply</property>
   </object>
   <object class="GtkImage" id="iconforward">
     <property name="visible">True</property>
-    <property name="can_focus">False</property>
-    <property name="margin_left">4</property>
+    <property name="can-focus">False</property>
+    <property name="margin-left">4</property>
     <property name="stock">gtk-go-forward</property>
   </object>
   <object class="GtkImage" id="iconscan">
     <property name="visible">True</property>
-    <property name="can_focus">False</property>
-    <property name="margin_left">5</property>
+    <property name="can-focus">False</property>
+    <property name="margin-left">5</property>
     <property name="stock">gtk-media-play</property>
   </object>
   <object class="GtkWindow" id="onboardingwindow">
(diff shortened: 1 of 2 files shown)


EmixamPP:
I tried to mimic you coding style. Feel free to tell me if there are something that you don't like.

boltgolt:
Depending on gnome-terminal is not ideal, this will break Howdy for many people (depends on X11 for example). Python can read from stdout and write to stdin, is there no way to make `linux-enable-ir-emitter` intractable in the background through the UI that way? Launching a terminal is not very user friendly.

I appreciate you working in my style though! Everything else looks great, i think the many changes in the onboarding UI XML (eg `can_focus` -> `can-focus`) are just a Glade upgrade?

EmixamPP:
> Depending on gnome-terminal is not ideal, this will break Howdy for many people (depends on X11 for example). Python can read from stdout and write to stdin, is there no way to make linux-enable-ir-emitter intractable in the background through the UI that way? Launching a terminal is not very user friendly.

Yes, I totaly agree with you. The problem is that `linux-enable-ir-emitter` requiere human interaction in order to generate a driver. I tried to implement a pipe to redirect input and output with `gi.Gio.Subprocess` and `gi.GLib`. Read the ouput can be done easily, but sync the whole system output/input/output/input/output... thought the GUI is an other story. The documentation of `gi.Gio.Subprocess` and `gi.Glib` (I personally think) is poor. And I didn't find a good topic on Internet about dealing with subprocess input and gi. 
So I need to work much more on this. 
I'll keep you in touch, but it may take a while.

> I appreciate you working in my style though! Everything else looks great, i think the many changes in the onboarding UI XML (eg can_focus -> can-focus) are just a Glade upgrade?

Hum, I used Glade from the Fedora repo, which is in version 3.38.2. And I just added one slide without touch to anything else.

EmixamPP:
If you prefer, I can download a specific version of Glade from their website.

EmixamPP:
Here is a better integration ! I had to change something in the way `linux-enable-ir-emitter` takes inputs but now everything is fine.

I have put on [this link](https://drive.emixam.be/d/s/mG7lVEVq7IRAw7tTEO5t18ya6APTMU2O/4T21E4LOfs8fZxpohxcDLnFvHojG3LE8-kbxgljT-EAk) two videos showing the case where the configuration fails and another where it works.

Feel free to edit (or add) the displayed messages.

boltgolt:
Thanks, changes look good! I like the interaction with LEIE a lot more this way. Small issue: `git` is not available by default on all systems, which is why either `curl` or `wget` (all supported systems have at least 1 of those) is used in the [debian installer](https://github.com/boltgolt/howdy/blob/beta/howdy/debian/postinst#L141). That's a small issue though and maybe i can just fix that one

I'm sorry i didn't have the time to look into and test it yet, i will do so ASAP

EmixamPP:
I added a condition in the commit https://github.com/boltgolt/howdy/pull/611/commits/67c776d5a4d8692b6c04f220160e881514d6b0ec to skip the camera configuration with `linux-enable-ir-emitter` if the user chooses a camera that is not infrared.

EmixamPP:
> Small issue: git is not available by default on all systems, which is why either curl or wget (all supported systems have at least 1 of those) is used in the debian installer. That's a small issue though and maybe i can just fix that one

I was using git for testing purposes, I was planning to change it to use a release and not git anyway (to avoid possible bugs). However, I'm still waiting to see if I need to change anything for its integration into Howdy before releasing a new version.

markbaas:
You are calling linux-enable-ir-emitter from command line. Both project are python. I'd rather see the python code from the other project being imported or merged into howdy.

EmixamPP:
Going through a command call greatly reduces the coupling between Howdy and linux-enable-ir-emitter.

I still have some plans for linux-enable-ir-emitter to better integrate it into the system, but it is beyond my current knowledge and so I will not be able to say what it will look like, or how it will work. 

And so directly calling the import the Python could break Howdy if linux-enable-ir-emitter is not at the same version as originally thought. 
It should also be noted that linux-enable-ir-emitter may have been installed previously and may also receive updates.

But indeed, we could include it into Howdy. But I'm not totally satisfied with it, like as said, I have some ideas for major improvements (although it works really great in its latest version).  

What bothers me with the implementation I made, is that it's quite dirty to install from Git a software, if the user uninstalls Howdy, this software will stay there, forgotten by the user. The best thing would be to make a PPA package for ditros derived from Debian. It's in my plans too, but I need to find some time to find out.

With hindsight, I have the impression that this graphic integration is perhaps not the best idea. A simple message that redirects to the GitHub page if the emitter doesn't work might be nice too.  

I don't know what you think @boltgolt ? I kind of feel like what I did was a bit of an ugly thing in terms of designing a program.

boltgolt:
Sorry for letting this one slip through, should have replied to this a while ago! I can add a step in the debian removal script to also automatically uninstall your tool, although that's not very clean either

EmixamPP:
The v4 of linux-enable-ir-emitter has been released, and I have published it in a PPA repository. I honestly think that the easiest and cleanest way to do this, would be to simply redirect them to the GitHub page that explains what to do. 

So I'm closing this PR, I'll reopen one in a few days with a simple integration in Howdy gtk that asks the user if the camera works or not.

Issue #624, opened 2021-12-17
Vertical side Feature request


neeks76:
Greeting to developer.

I'm very thankful to this solution.
It's good work in horizontal.

I would like to request a feature is vertical side.
When I make notebook to vertical from horizontal, then the howdy can't work.

I tried to take a picture face on horizontal, but this solution not expected.

Plz, solve or new feature release.

boltgolt:
Could be implemented quite easily, but would really slow down detection if enabled

matan-arnon:
Will be trying to implement this as my first open source contribution.

boltgolt:
That would be very appreciated! I was thinking of adding a config option that allows you to specify extra angles Howdy should run face detection on, like: 

```
rotation_angles = 0,90 
```

Which would just simply rotate every other frame 90 degrees

chenxiex:
I'm using howdy beta on Arch Linux with [https://aur.archlinux.org/packages/howdy-beta-git](https://aur.archlinux.org/packages/howdy-beta-git). The vertical side verification works with howdy pam, but howdy test doesn't show the rotated frames, nor does it show the green circle.

PR #627, opened 2021-12-31
Vertical login

matan-arnon:
This allows logging in when notebook is held vertically as requested in #624 
Uses xlib to get screen dimensions (allowing to assume what orientation screen is in).
If screen is portrait instead of standard landscape, alternates rotating images 90 degrees clockwise and counter-clockwise before scanning for face.


howdy/debian/postinst
```diff
@@ -118,6 +118,11 @@ log("Upgrading numpy to the latest version")
 # Update numpy
 handleStatus(subprocess.call(["pip3", "install", "--upgrade", "numpy"]))
 
+log("Installing xlib")
+
+# Install xlib
+handleStatus(subprocess.call(["pip3", "install", "--upgrade", "python3-xlib"]))
```
boltgolt, line 124: This feels like an external dependency for something relatively few users are going to use, maybe we can make it a config option and ask users to install this themselves? That would also remove a point of failure in installation

howdy/src/compare.py
```diff
@@ -22,6 +22,7 @@
 import snapshot
 import numpy as np
 import _thread as thread
+import Xlib.display as display
```
boltgolt, line 25: What happens when a user is running Wayland?

howdy/src/compare.py
```diff
@@ -193,9 +194,19 @@ def send_to_ui(type, message):
...
 max_height = config.getfloat("video", "max_height", fallback=0.0)
-# Get the height of the image
-height = video_capture.internal.get(cv2.CAP_PROP_FRAME_HEIGHT) or 1
 
+# Get screen orientation
+landscape = True
+dsp = display.Display()
+screen_width = dsp.screen().width_in_pixels
+screen_height = dsp.screen().height_in_pixels
+if screen_height > screen_width:
```
boltgolt, line 203: I'm worried this will trigger for people using external vertical monitors, i really think this should be disabled by config option by default

boltgolt (changes requested):
Thank you very much for these changes! This will help quite a few people i think. However, the dependency on X server and the possibilities of false positives worry me a bit. Maybe reading from a config value and only then importing anything X related would be better?

matan-arnon:
@boltgolt, makes sense to me. Could add it to the config file and then instead of importing X, just automatically rotate the images when checking for faces. Would make authentication slower for the few people that use the feature but would eliminate the X dependency.
What do you think?

boltgolt:
Hm yes that would also work and be a bit simpler to implement, but would be a shame to throw some of your work away. We could just tell people in install the dependency in the config:

```ini
# Rotate captured frames so faces are upright
# Can be either false, auto, or an array of angles like 0,90,180
# Please install the required dependency before enabling auto mode: 
#  pip3 install python3-xlib
rotate = auto
```

You don't have to implement the array part at all of course i'm just thinking out loud here

boltgolt (approved):

boltgolt:
Thank you so much! Will be released together with the other beta v3 changes

PR #647, opened 2022-04-05
 Fix :"Unknown error: 1" on gnome-screensaver

calvinweb:


boltgolt:
Interesting patch, could you describe what it does and why? Emptying $PATH is a very large step to take

calvinweb:
> Interesting patch, could you describe what it does and why? Emptying $PATH is a very large step to take

because some librarys(eg. different verision of dlib and opencv) may make it dosn't work

boltgolt:
I can see why this fixes that, but i would anticipate that for others this change would break their previously working setup. Unfortunately i'm going to have to decline this change but thanks for submitting it!

calvinweb:
but it only effect in python life cycle ,pleace open it again

boltgolt:
But what if people use pyenv to make sure Howdy runs in python3? Clearing the $PATH will completely break Howdy for them.
Additionally, the check that the path starts with /home seems completely arbitrary to me

calvinweb:
it isn't environment var,it only a var in python runtime

calvinweb:
but i think this is a mistake from python,python should find libraries from system paths at first

PR #654, opened 2022-04-21
Just adding some linguistic english niceties to readme

ryancastro:


boltgolt:
Small change but appreciated, thanks!

Issue #671, opened 2022-05-20
Python Exception "AttributeError" on /usr/lib/security/howdy/cli/test.py


NP-chaonay:
When using value of gamma that not "-1" in conifguration file given from "howdy config" command

and doing command "howdy test" got this:

```
Traceback (most recent call last):
  File "/usr/local/bin/howdy", line 109, in <module>
    import cli.test
  File "/usr/lib/security/howdy/cli/test.py", line 194, in <module>
    video_capture.intenal.set(cv2.CAP_PROP_AUTO_EXPOSURE, 1.0)  # 1 = Manual
AttributeError: 'VideoCapture' object has no attribute 'intenal'. Did you mean: 'internal'?
```

I've searched for similar issues already, and my issue has not been reported yet.

Linux distribution (if applicable): Ubuntu 22.04 LTS

Howdy version (`sudo howdy version`): Howdy 2.6.1
Howdy from official repo, I have looked through beta/stable/dev branch seem to be no fixed yet.

NP-chaonay:
changing intenal to internal as following exception description fixed this issues.

NP-chaonay:
changed to pull request instead. (https://github.com/boltgolt/howdy/pull/672)

boltgolt:
Thought we fixed this typo everywhere, thanks for the PR!

PR #672, opened 2022-05-20
rename wrong attribute name to prevent AttributeError Python Exception

NP-chaonay:
according to https://github.com/boltgolt/howdy/issues/671

also pls applied on both beta/dev branch

### DIFF
--- a/src/cli/test.py
+++ b/src/cli/test.py
@@ -192,8 +192,8 @@ try:
 			# are captured and even after a delay it does not
 			# always work. Setting exposure at every frame is
 			# reliable though.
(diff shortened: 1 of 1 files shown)


boltgolt:
Thank you! The file structure of the project is different now on the target beta branch. Could you fix the conflict?

NP-chaonay:
> Thank you! The file structure of the project is different now on the target beta branch. Could you fix the conflict?

It seem that I use origin of branch "dev", so it have problem when PR to target branch "beta".
in that case I think new PR is required or instead editing code manually.

I'm not fond of GitHub too much btw.

boltgolt:
Beta is the current target and default branch unfortunately, but you should be able to rebase? Otherwise opening a new PR is not a problem at all

NP-chaonay:
Move to #673

PR #673, opened 2022-05-25
rename wrong attribute name to prevent AttributeError PYthon Exception

NP-chaonay:
Same as  #672 but different head repo branch


boltgolt:
Great, thanks!

PR #683, opened 2022-06-30
linux-enable-ir-emitter

EmixamPP:
Improvement :
- ~identify devices using their `/dev/video` path, instead `/dev/v4l/by-path` since this last one is a simlink~

New :
- slide 3 :  
   1. trigger the emitter (if an ir camera is selected)
   2. the user has to answer if the emitter works or not
   3. if yes, display slide 4
   4. if no, display linux-enable-ir-emitter GitHub link (which is clickable)
  
 Unknown issue observed : 
 - sometimes after having click on `next`, the next slide is not directly displayed
 - on the slide 2 if we click on the `next` button when it is not enabled, the clicks are registered and triggered when it will be activated

### DIFF
--- a/howdy-gtk/src/onboarding.glade
+++ b/howdy-gtk/src/onboarding.glade
@@ -1,56 +1,56 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.38.2 -->
 <interface>
   <requires lib="gtk+" version="3.12"/>
   <object class="GtkImage" id="iconcancel">
     <property name="visible">True</property>
-    <property name="can_focus">False</property>
-    <property name="margin_right">4</property>
+    <property name="can-focus">False</property>
+    <property name="margin-right">4</property>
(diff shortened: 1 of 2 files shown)


boltgolt:
Thank you for these changes! I like all of them with the exception of the camera path change. 

The reason Howdy uses /by-path is that these paths are consistent during reboot. Using a videoX device does not guarantee that that specific hardware will be used on the next boot, as these numbers are assigned during boot time and can swap. My IR camera is video1 most of the time but sometimes switches to video0. 

Additionally, using the path means that the fake-webcam attack vector is harder to execute. In this attack an unauthorised person would disconnect the USB webcam and attach a fake USB webcam that simply exposes a static picture of the victim as a V4L device. This is much harder to execute if we are using /by-path as the attacker would also have to clone the USB signature of the original camera.

EmixamPP:
Let me know if I can do something else for this PR (don't worry if you haven't had time to look yet)

boltgolt:
No issues with the PR, just me not dedicating the time to it, thank you for the changes!

Issue #630, opened 2022-01-13
Integration of howdy with gnome polkit doesn't work


Itotutona:
Hello,

I tryed to add howdy to polkit but it doesn't work, could you help me please ?

> /etc/pam.d/polkit-1

```#%PAM-1.0
auth       sufficient   pam_python.so /lib64/security/howdy/pam.py
auth       include      system-auth
account    include      system-auth
password   include      system-auth
session    include      system-auth
```

Linux distribution: Fedora 35

Howdy version: 2.6.1

ad-on-is:
Just installed  howdy with GNOME 41.3 and it works.

Although, my line looks different than yours.

```
auth       sufficient   pam_python.so /lib/security/howdy/pam.py
```

`lib` instead of `lib64`

Itotutona:
Are you using Fedora ? I use /lib64/security/howdy/pam.py because I follow the documentation in the copr

ad-on-is:
I'm on EndeavourOS, running vanilla GNOME

Itotutona:
My problem is I never see polkit. My webcam is active (flash is active) but I don't see any popup (by default polkit open a popup). And until I kill the application, my camera is active but I can't connect.

ad-on-is:
Aaah, did you check the howdy config.ini. There's some setting about showing a notification.

ad-on-is:
And btw, your title is misleading, since everyone would assume it's not working at all. Just try to be more precise.

ad-on-is:
Sorry, my bad, didn't know how Fedora handles it.

nickavem:
I've been troubleshooting this, and it doesn't seem to be setting off any SELinux alarms. Getting same behavior, though.

Itotutona:
I works on [Arch Linux](https://archlinux.org) with the howdy package.

EmixamPP:
I can confirm that this is not due to SELinux.

Then I investigated a bit, and it appears that when authenticating with polkit, it "spam" the authentication request and kill the previous process and so it never gets the response from Howdy which was comparing the face read by the camera to the model. 

If I modify the pam.py script that handles pam, to reply that the authentication is approved before calling Howdy, then there is no problem, pam doesn't call the authentication script again.

So the problem lies in the polkit configuration.

EmixamPP:
[...]

PR #687, opened 2022-07-15
Fix Fedora Polkit

EmixamPP:
This is a fix for the issue https://github.com/boltgolt/howdy/issues/630

If think that on Fedora, polkit detect that the subprocess `compare.py` is created and to avoid security issue it is killed. 
However, if we run this subprocess through a shell instance, this is no more a problem.
So this fix is not really clean, but it works. 

In order to provide a proper solution, I think we need to investigate polkit to create a policy or something like that. But personally, I don't know anything about it, and this seems complex. It is also possible that there is no such thing and that it is simply a bug.

---
I made the modification from the tag of the current version of Howdy: v2.6.1.

In order to provide this fix for Fedora users we have several choices:

* An official version of Howdy 2.6.2 is created
* The Fedora's repo maintainer creates a "hot fix", but we keep the v2.6.1 tag as source. The version on Fedora would become 2.6.1-7 (currently 2.6.1-6)
* I create a tag on the `fedora-polkit` branch of my fork, which is temporarily used as source. The version on Fedora would become 2.6.1-7
* We do nothing for Fedora at the moment, and we wait for the next official release, the v3.0.0, which rewrites the pam module in C++

### DIFF
--- a/src/pam.py
+++ b/src/pam.py
@@ -41,7 +41,7 @@ def doAuth(pamh):
 	syslog.syslog(syslog.LOG_INFO, "Attempting facial authentication for user " + pamh.get_user())
 
 	# Run compare as python3 subprocess to circumvent python version and import issues
-	status = subprocess.call(["/usr/bin/python3", os.path.dirname(os.path.abspath(__file__)) + "/compare.py", pamh.get_user()])
+	status = subprocess.call(["/usr/bin/python3 " + os.path.dirname(os.path.abspath(__file__)) + "/compare.py " + pamh.get_user() + " &> /dev/null"], shell=True)
 
 	# Status 10 means we couldn't find any face models
 	if status == 10:


EmixamPP:
@principis What do you think of that?

validatedev:
Works flawlessly, can confirm from Fedora 36. Thanks so much!

mikecharles:
So, do I only need to modify `/lib64/security/howdy/pam.py`? Do I have to restart or do anything else for this to start working?

EmixamPP:
Yes, just change the line 44 of `/lib64/security/howdy/pam.py` with this one :
```python
status = subprocess.call(["/usr/bin/python3 " + os.path.dirname(os.path.abspath(__file__)) + "/compare.py " + pamh.get_user() + " &> /dev/null"], shell=True)
```

Okazakee:
this is fantastic, thanks, +1 for merging or copr repo hotfix

wheelerlaw:
I can confirm that this also worked for my installation. It also fixed the issue with being prompted to enter my password to unlock the Gnome keychain _after_ I had logged in with Howdy

principis:
I'll add this in the copr repo.

wheelerlaw:
Is polkit exclusive to Fedora, or is it a Gnome thing?

boltgolt:
The next release of Howdy will completely remove pam.py to move over to a compiled pam_howdy.so and drop the dependency on pam-python. This file is now gone on the beta branch. Is it worth merging this into master still?

EmixamPP:
@wheelerlaw A Gnome thing

boltgolt:
Closing now because of me being very very slow with updating Howdy, and moving forward with the beta branch deprecates this

PR #690, opened 2022-07-20
Update Debian build files

musikid:
Add the necessary build dependencies and remove deprecated python requirement.
Also use `install` instead of `compile` for meson command, since the version shipped with Ubuntu 20.04 doesn't support it.
Not sure about the locales though, should we keep them?

### DIFF
--- a/howdy/debian/rules
+++ b/howdy/debian/rules
@@ -11,7 +11,7 @@ build:
 	# Create build dir
 	meson setup -Dinih:with_INIReader=true build src/pam
 	# Compile shared object
-	meson install -C build
+	ninja -C build
 
 clean:
 	# Delete mason build directory
--- a/howdy/debian/rules
+++ b/howdy/debian/rules
@@ -11,12 +11,10 @@ build:
 	# Create build dir
 	meson setup -Dinih:with_INIReader=true build src/pam
 	# Compile shared object
-	meson compile -C build
+	meson install -C build
 
 clean:
 	# Delete mason build directory
 	rm -rf ./build
 	# Force remove temp debian build directory
 	rm -rf ./debian/howdy
-	# Make sure subprojects get pulled locally
-	meson subprojects download --sourcedir src/pam
--- a/howdy/debian/control
+++ b/howdy/debian/control
@@ -2,7 +2,7 @@ Source: howdy
 Section: misc
 Priority: optional
 Standards-Version: 3.9.7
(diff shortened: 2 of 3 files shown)


howdy/debian/rules
```diff
@@ -11,12 +11,10 @@ build:
 	# Create build dir
 	meson setup -Dinih:with_INIReader=true build src/pam
 	# Compile shared object
-	meson compile -C build
+	meson install -C build
```
eli-schwartz, line 14: Install will actually attempt to write files to `/lib/security`, so this will not work. Trying to run an unrelated command because it attempts to build as a side effect is wrong.

Keep in mind that `meson compile -C build/` is a wrapper script that does *nothing* other than run `ninja -C build/` for you. The only reason `meson compile` exists at all is for portability to Windows, since it may then run `msbuild` instead of `ninja` (msbuild invokes the Visual Studio solution files).

You don't care about that. :) Just use ninja.

eli-schwartz (changes requested):

howdy/debian/rules
musikid: Oh yes, sorry, I had forgotten that it was a wrapper for `ninja` actually!

boltgolt:
Thank you! One thing, removing `meson subprojects download` seems to me like it would break the launchpad build process. The issue with launchpad is that it has to be build completely offline so the build is deterministic: The same source will always build to the same binary on the same machine. Is there a way to still pull the inih subproject locally?

musikid:
It will be pulled as a subproject only if it wasn't found already on the system. Since I added `libinih-dev` to the build dependencies, it will no longer require downloading it as a subproject and therefore use the one installed.

boltgolt:
Thank you! I've put your changes in the debian files but can't merge due to conflicts there

musikid:
No problem, tell me if you need any help!

boltgolt:
@musikid i actually ran into an issue that my C++ skills are not good enough to handle: 

In the current beta code, when Howdy errors out the PAM module will also terminate the password prompt. This is an issue because anyone without a model configured will get status code 10, which will terminate both Howdy and prompt, making it impossible to authorize. In effect this makes it impossible to access sudo as soon as you install Howdy which is not really the goal

My preferred solution would be to wait for `pass_task` to finish if Howdy gives any status code that is not 0, but as said i can't get that to work here

musikid:
How can I reproduce this issue? Because it seems to work on my side.

musikid:
How can I reproduce this issue? Because it seems to work on my side.

PR #691, opened 2022-07-25
Fix config.getboolean

wmyfelix:


boltgolt:
Thanks for the patch! pam.py is no longer used now that howdy is switching over to its own pam_howdy.so file and dropping the pam-python dependency fully. To clear the confusion i've deleted pam.py from the beta branch now

PR #692, opened 2022-07-26
Try 4 times more if can't read a frame.

mingww64:
I'm using a webcam, in my scripts, it will start and stop with howdy, but the connection has a delay so howdy can't capture frames forthwith. This won't impact the efficiency but improve the stability, so I'm considering whether l shall push it.

### DIFF
--- a/howdy/src/recorders/video_capture.py
+++ b/howdy/src/recorders/video_capture.py
@@ -3,6 +3,7 @@
 
 # Import required modules
 import configparser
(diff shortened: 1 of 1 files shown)


boltgolt:
I'm not really sure about this PR, wouldn't a configurable delay be easier?

felicia-wen:
Maybe this one is simpler to users and more flexible on different camera?

PR #709, opened 2022-09-28
Add python3 pam-python support

mikacousin:
Hi,
As a pam-python based on python3 is available on Archlinux (https://aur.archlinux.org/packages/pam-python-git, based on https://github.com/castlabs/pam-python), howdy need to import good configparser module.

### DIFF
--- a/howdy/src/pam.py
+++ b/howdy/src/pam.py
@@ -4,13 +4,17 @@
 import subprocess
 import os
 import glob
+import sys
 import syslog
 
-# pam-python is running python 2, so we use the old module here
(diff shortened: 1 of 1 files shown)


mikacousin:
> also remove dependencies for python2. for it's out of support.

I think it must be in another PR once this one gets merged.
And maybe you need to be sure that a pam-python module based on python3 is available on the different distributions.

mikacousin:
Sorry, I didn't see that this branch no longer uses pam-python.
I can do another PR on the dev branch if you want.

PR #719, opened 2022-10-31
Add python3 pam-python support

mikacousin:
As in https://github.com/boltgolt/howdy/pull/709 I don't use the good branch, this is another attempt to use pam-python based on python3... With the good branch :-)
Tested on Archlinux.


boltgolt:
Thanks, merged into dev for older versions!

PR #736, opened 2022-12-08
[beta] Add pam_python3 support and use updated config variable names

jezek:
For the `auth sufficient pam_python3.so /lib/security/howdy/pam.py` rule work in `/etc/pam.d/` files with `Howdy 3.0.0 BETA`, some fixes are needen in the some `.py` files.

First, the `configparser` initialization in `pam.py` needs to be tweaked for python3. This is already done in PR #719, so I merged it to this PR.

Second. There were some changes in config variable names. These changes have to be implemented to the `.py` files. Some effort was done in PR #691, which was merged. All other variable name updates were added by me in the last commit 0803465cd92ffd3bbcf20139d2f06f979f302a52.

I've also spotted, that the variables `capture_failed` and `capture_successful`, which changed to `save_failed` and `save_successful`, were not added to `debian/postins` upgrade block. So I took the liberty to add them there.


boltgolt:
In the beta pam.py is no longer used, and the file is now gone from the repo. Thanks so much for the other changes!

PR #762, opened 2023-03-06
config: add fallbacks

### LINKED ISSUE #761: Howdy cannot use an incomplete config.ini
I'm writing a [NixOS module for Howdy](https://github.com/NixOS/nixpkgs/pull/216245), and I'm letting the user define the Howdy config as a nix expression that gets translated to ini and placed in `/etc/howdy/config.ini`. While doing this, I have looked through the code and found that fallbacks are set, so I assumed they were for values that do not appear in the config file. However, when using this `config.ini`:

```ini
[core]
no_confirmation=true
[debug]
end_report=true
[video]
dark_threshold=90
```

Howdy gives me errors:
```py
sudo -E howdy test
  File "/nix/store/iw1vmh509hcbby8dbpsaanbri4zsq7dj-python3-3.10.10/lib/python3.10/configparser.py", line 791, in get
    value = d[option]
  File "/nix/store/iw1vmh509hcbby8dbpsaanbri4zsq7dj-python3-3.10.10/lib/python3.10/collections/__init__.py", line 986, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/nix/store/iw1vmh509hcbby8dbpsaanbri4zsq7dj-python3-3.10.10/lib/python3.10/collections/__init__.py", line 978, in __missing__
    raise KeyError(key)
KeyError: 'recording_plugin'

During handling of the above exception, another exception occurred:

  File "/run/current-system/sw/bin/howdy", line 120, in <module>
    import cli.test
  File "/nix/store/yiyrf0igvkax54spw0lqpgrwfr5gyq8g-howdy-unstable-2023-02-28/lib/security/howdy/cli/test.py", line 24, in <module>
    if config.get("video", "recording_plugin") != "opencv":
  File "/nix/store/iw1vmh509hcbby8dbpsaanbri4zsq7dj-python3-3.10.10/lib/python3.10/configparser.py", line 794, in get
    raise NoOptionError(option, section)
configparser.NoOptionError: No option 'recording_plugin' in section: 'video'
```

Is it possible to have an incomplete config file and have Howdy take care of the missing options?

fufexan:
Add missing fallbacks for config options, based on values in `config.ini`.
This allows using a config file where only some of the keys are set.
Fixes #761.

Untested, will add an edit once I've finished testing.
EDIT: auth seems to be working everywhere except `swaylock`. However, if a file with every key set is present, it will work properly.

### DIFF
--- a/howdy/src/cli/disable.py
+++ b/howdy/src/cli/disable.py
@@ -35,13 +35,13 @@ else:
 	sys.exit(1)
 
 # Don't do anything when the state is already the requested one
-if out_value == config.get("core", "disabled"):
+if out_value == config.get("core", "disabled", fallback=True):
 	print(_("The disable option has already been set to ") + out_value)
 	sys.exit(1)
 
 # Loop though the config file and only replace the line containing the disable config
 for line in fileinput.input([config_path], inplace=1):
-	print(line.replace("disabled = " + config.get("core", "disabled"), "disabled = " + out_value), end="")
+	print(line.replace("disabled = " + config.get("core", "disabled", fallback=True), "disabled = " + out_value), end="")
 
 # Print what we just did
 if out_value == "true":
--- a/howdy/src/recorders/video_capture.py
+++ b/howdy/src/recorders/video_capture.py
@@ -33,7 +33,7 @@ class VideoCapture:
 
 		# Check device path
 		if not os.path.exists(self.config.get("video", "device_path")):
-			if self.config.getboolean("video", "warn_no_device"):
+			if self.config.getboolean("video", "warn_no_device", fallback=True):
 				print(_("Howdy could not find a camera device at the path specified in the config file."))
 				print(_("It is very likely that the path is not configured correctly, please edit the 'device_path' config value by running:"))
 				print("\n\tsudo howdy config\n")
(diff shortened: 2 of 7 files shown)


howdy/src/cli/disable.py
```diff
@@ -35,13 +35,13 @@
 	sys.exit(1)
 
 # Don't do anything when the state is already the requested one
-if out_value == config.get("core", "disabled"):
+if out_value == config.get("core", "disabled", fallback=False):
```
boltgolt, line 38: I would like the fallbacks to be as safe as possible, would it be okay for your implementation if this defaulted to to true? Then howdy would be disabled automatically

howdy/src/recorders/video_capture.py
```diff
@@ -100,21 +100,22 @@ def _create_reader(self):
 		"""
 		Sets up the video reader instance
 		"""
+		recording_plugin = self.config.get("video", "recording_plugin", fallback="opencv")
```
boltgolt, line 103: Good spot!

boltgolt (changes requested):

howdy/src/cli/disable.py
fufexan: Sure.

boltgolt (approved):

boltgolt:
Awesome, thank you!

fufexan:
FYI: this does not solve the original issue completely. While howdy does not crash anymore, it still will not work in certain programs, such as sddm and swaylock.

zaolin:
Any insight why @fufexan ?

boltgolt:
That doesn't sound like it's caused by the same issue

fufexan:
@zaolin I just forgot to add `no_confirmation = true` in the config file. That fixed it.

PR #773, opened 2023-03-30
Put all path variables into a separate module.

arrowd:
This makes it easier for downstream packagers to customize where howdy installs its files.

### DIFF
--- a/howdy/src/cli/add.py
+++ b/howdy/src/cli/add.py
@@ -8,6 +8,7 @@ import json
 import configparser
 import builtins
 import numpy as np
+import paths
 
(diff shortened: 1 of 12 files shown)


arrowd:
Can we get this merged in?

GMasucci-Twig (approved):

musikid (commented):
We also need to propagate these paths to the PAM module, probably by using `configure_file` with Meson.
On another note, wouldn't it be better to use path manipulation functions instead of concatenation?

boltgolt:
Superseded by #803, thanks for the input!

PR #775, opened 2023-04-04
Adding missing `v4l-utils` package to Ubuntu dependencies

raelgc:
Howdy missed my camera device during install because the setup process relies on `v4l`, which was not installed.


boltgolt:
Thanks for the patch! Howdy works without v4l if you configure your camera manually and choose a different recorder. Maybe it would be better as a recommended package?

raelgc:
@boltgolt Indeed, it makes more sense. I just updated the PR :)

boltgolt:
Perfect, thank you!

Issue #769, opened 2023-03-25
Support for `doas`


hackdev17:
With `doas` I was not able to run the /usr/bin/howdy even as a privileged user and it is same with `root`

```
[user@host ~]$ doas howdy test
doas (user@host) password: 
Please run this command as root:

        sudo howdy test
[user@host ~]$ su -
Password: 
[root@host ~]# howdy test
Please run this command as root:

        sudo howdy test
``` 
Suggest any configuration changes if required

More info
```
OS : Arch Linux

Package information
Name            : howdy
Version         : 2.6.1-2
Description     : Windows Hello for Linux
Architecture    : x86_64
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : None

Files related to the package
[user@host ~]$ doas ls -R / | grep howdy
doas (user@host) password: 
~/Downloads/howdy
~/Downloads/howdy/dlib-models
~/Downloads/howdy/dlib-models/age-predictor
~/Downloads/howdy/dlib-models/gender-classifier
/usr/lib/security/howdy
/usr/lib/security/howdy/cli
/usr/lib/security/howdy/dlib-data
/usr/lib/security/howdy/pam-config
/usr/lib/security/howdy/recorders
/usr/share/licenses/howdy
/var/lib/pacman/local/howdy-2.6.1-2
```

boltgolt:
That's weird because howdy checks if .geteuid() returns 0. What is your root user id?

hackdev17:
As a normal user my id is 1000

hackdev17:
I have tested `howdy` with `sudo` it works fine though, any file that invoke `sudo` I havent found yet, where do the `sudo` get invoked in the process `sudo howdy test`

boltgolt:
Sudo doesn't get invoked, in a simplified way: By putting `sudo` before a command that command is executed as if the root user has executed it, so the user id of the process is 0

What does `doas id` give you?

hackdev17:
```
[user@host ~]$ doas id
doas (user@host) password: 
uid=0(root) gid=0(root) groups=0(root)
```

hackdev17:
The issue has now been resolved !
Following this guide

I have added the line to `/etc/pam.d/doas`
`auth sufficient pam_python.so /lib/security/howdy/pam.py`

```
[user@host ~]$ doas chmod -R 755 /lib/security/howdy
[user@host ~]$ cat /etc/pam.d/doas 
#%PAM-1.0
[...]

PR #798, opened 2023-06-07
Add doas support

### LINKED ISSUE #642: Accept Doas as an alternative to Sudo
Even with root privileges howdy doesn't let us use its commands unless you use sudo. Adding a variable in the config file for it would be great. It would also be helpful to show the config path when you get the message about needing elevated permissions.

izvyk:
Resolves #642 and #769 

I've added `doas` support and rewritten the part that is responsible for the determination of the user.


boltgolt:
Thank you!

PR #802, opened 2023-06-19
Ignore compare error if password is typed

musikid:


boltgolt:
Awesome! Thanks once again

PR #817, opened 2023-07-22
Replaced deprecated python-pam dependency.

SotchNam:
Changed the dependency to use the new python-pam package  from extra repo instead of the old one.


boltgolt:
Thanks for your work, but take a look at the `beta` branch, there's work on a version that does not use put him for the direct PAM at all

SotchNam:
Thanks, if that is the case then please remove the dependency from the pkgbuild.

PR #828, opened 2023-08-21
Fix some bugs

ciaoly:
I found it doesn't import the os module in preinst script, thus I fixed it.
While I installed the beta version pack and run `howdy disable 1`, I got a `FileNotFoundError` because the lacking of slash while calling function `os.path.dirname`. I don't know if it is a bug. But after this commit I can run it properly.


boltgolt:
Thank you!

PR #831, opened 2023-08-28
Ensure Model ID is unique

Gliese852:
The ID of the newly created model was obtained as the length of the array of models, but if some models "in the middle" were deleted, the new ID could be the same as the existing one.

We are sure that the last ID is the maximum, so we just take +1 than the last element.

UPDATE: Sorry, posted an unsigned commit first.


boltgolt:
Good catch, thank you so much!

PR #834, opened 2023-08-30
Add configurable device frame rate

Joseph-DiGiovanni:
Currently OpenCV defaults to the first detected (and sometimes lowest) frame rate. This is bad enough for detection on its own, but some IR emitters do not function properly at all frame rates the camera may support (like my surface laptop studio). I've added a device_fps option to the config to set this manually, as well as use the highest frame rate by default on new installs.


boltgolt:
Great idea, thanks!

PR #837, opened 2023-09-13
Add option to make howdy inactive during night (dark)

tokox:
I added option which will make howdy inactive during night (on specified hours in pam module and on detection in compare.py). This will allow users to not have to wait for howdy timeout when there is no way howdy would work.

### DIFF
--- a/howdy/src/compare.py
+++ b/howdy/src/compare.py
@@ -15,6 +15,7 @@ import sys
 import os
 import json
 import configparser
+import re
 import dlib
 import cv2
 import datetime
@@ -23,9 +24,14 @@ import subprocess
 import snapshot
 import numpy as np
 import _thread as thread
+import syslog
+import time
 
 # Allow imports from the local howdy folder
 sys.path.append('/lib/security/howdy')
+
+# Open syslog
+syslog.openlog("pam_howdy", 0, syslog.LOG_AUTHPRIV)
 
 from recorders.video_capture import VideoCapture
 from i18n import _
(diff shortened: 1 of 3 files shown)


tokox:
I have only normal camera, which is not working properly at night. This won't affect people with IR emitters, but could help people with day-only webcams.

boltgolt:
Thank you for putting in the effort to open a PR for this! 
Howdy is primarily made for IR cameras, and i think this is a pretty big patch for pretty niche functionality. In addition, where i live the time it gets dark varies a lot during the year, so i would have to manually update the time values every month or so for it to stay accurate

Maybe this would be better as two cron jobs, one that runs `sudo howdy disable 1` at sundown and one that re-enables it at dawn?

tokox:
I don't know what you mean by "a lot", but I don't think it would be an issue with 'camera' option and hours set with margin. 'on' options would not work anyway, because of light which can be turned on at any time.

I also have one question: when 3.0.0 version?

boltgolt:
[Sunset somewhere between +/- 16:00 in winter and +/- 22:00 in summer](https://www.timeanddate.com/sun/denmark/copenhagen)

To be honest i think almost everything is ready for 3.0.0, i don't have a great overview of what still needs to be done

tokox:
> [Sunset somewhere between +/- 16:00 in winter and +/- 22:00 in summer](https://www.timeanddate.com/sun/denmark/copenhagen)

Very sililar to me.

> To be honest i think almost everything is ready for 3.0.0, i don't have a great overview of what still needs to be done

I am waiting for it.
Thank You!

PR #853, opened 2023-11-07
Connect to signals of the shown window

Gliese852:
In the case of `OnboardingWindow` and `MainWindow`, the class itself is not shown as a window, but its `window` field is shown, so when the window is closed, the application does not terminate - the wrong signals is connected to the handler.


Gliese852:
I corrected it a little, because it is not always called with all the args. (An error occurred when clicking the "Cancel" button in the wizard.)

boltgolt:
Thank you!

PR #855, opened 2023-11-09
Fix segfault when adding a model via howdy-gtk

Gliese852:
On all test machines available to me (3), the `map` event was apparently fired _before_ the "Please look directly into the camera" dialog was finally created. Therefore, the dialog was not displayed at all when scanning a face, and a segfault occurred when calling destroy().

Copied the code from onboarding.py, it works well there.


boltgolt:
Thank you!

PR #857, opened 2023-11-15
Fix things regarding translations

Gliese852:
- do not translate parser arguments, because these are the keys in the `args`. And therefore, for example, `args.command` will be missing further.

  - translate button names in `main.glade`


boltgolt:
Thank you!

PR #871, opened 2023-12-20
fix errors when install_in_site_package in meson.build.

xiangluoming:
According to https://github.com/mesonbuild/meson/issues/12601 , py.install_sources shouldn't use install_mode.


musikid:
Thanks, it worked previously but seems that it shouldn't have been documented as such.

Issue #735, opened 2022-12-08
[beta] Failed to send Enter press


jezek:
Hello,

when I add `auth sufficient pam_howdy.so` into `/etc/pam.d/polkit-1` and use some program requiring admin rights (e.g. `pamac-gtk`), the authentication window shows up, the IR camera diodes are flashing, it seems the face is recognized, but the authentication window writes an `Failed to send Enter press, waiting for user to press it insted` error and it is not possible to authenticate or send the Enter press via keyboard. Only cancel works.

![Screenshot from 2022-12-07 18-14-27](https://user-images.githubusercontent.com/5408235/206521387-87673aa4-5dac-4c79-aa9f-d0137876864c.png)

Is there any way to send the Enter press which I don't know? Or is it possible to tweak something in config to make it work?

----

I've searched for similar issues already, and my issue has not been reported yet.

Linux distribution (if applicable): Manjaro Linux x86_64, GNOME 43.1, wayland

Howdy version (`sudo howdy version`): Howdy 3.0.0 BETA

---

PS: The `auth sufficient pam_howdy.so` works as expected for `/etc/pam.d/sudo` and `/etc/pam.d/system-local-login`. Thank you.

boltgolt:
Thanks for looking at the beta, have you tried pressing any key and then enter? The dialog might not allow empty passwords

jezek:
@boltgolt This issue is kind of sorted for me. Shall I close this issue or leave it open until it is properly handled with `pam_howdy.so`? Or is it impossible to implement for now?

PR #888, opened 2024-02-08
Add text to avoid empty input for input workaround

soraxas:
For the `input` workaround, we add a space key event to input some text before pressing the enter key. This avoid the error that happens due to gtk preventing empty input.

This make it so that the *workaround* works in gtk auth like:

Screenshot from 2024-02-06 17-18-11

Pretty much after the original workaround doesn't work, we will enter a space key before pressing enter


soraxas:
closes #735

saidsay-so (changes requested):

PR #903, opened 2024-04-10
fix(pam): use environ variable when getenv doesn't work

### LINKED ISSUE #902: SSH_CONNECTION  security issue is back
like #74 

use aur howdy-beta-git, The same thing happened: "Face recognition on login via ssh #74"

saidsay-so:
On some programs like sudo, getenv doesn't return the environment variable value.
We check the `char **environ` variable also to not miss those values.

Close #902

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ To install them on Debian/Ubuntu for example:
 sudo apt-get update && sudo apt-get install -y \
 python3 python3-pip python3-setuptools python3-wheel \
 cmake make build-essential \
-libpam0g-dev libinih-dev libevdev-dev \
+libpam0g-dev libinih-dev libevdev-dev python3-opencv \
 python3-dev libopencv-dev
 ```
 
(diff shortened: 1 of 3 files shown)


yuhldr:
I found a new problem, gnome-remote-desktop 46.0-2, when other computers remotely control A by xrdp, A needs to unlock the screen or elevate privileges (such as modifying the printer), A will give priority to howdy face unlock

Screenshot from 2024-04-10 17-55-20

principis:
This doesn't really make sense to me. Why is it that `environ` may contain variables that `getenv` doesn't return? It should be the same.

saidsay-so:
> This doesn't really make sense to me. Why is it that `environ` may contain variables that `getenv` doesn't return? It should be the same.

Frankly I don't know myself, I wasn't thinking that it was possible but I was able to reproduce the bug as it was initially reported and the fix works.

DragoonZ:
when will this be merged?

yuhldr:
I hope this be merged

boltgolt:
Thank you once again @saidsay-so!

Issue #1000, opened 2025-03-09
Default editor not respected


rdong8:
_Please describe the issue in as much detail as possible, including any errors and traces._
_If your issue is a camera issue, be sure to also post the image generated by running `sudo howdy snapshot`._

[This](https://github.com/boltgolt/howdy/blob/beta/howdy/src/cli/config.py#L23) line causes an exception when `EDITOR` is not set to `/bin/nano`:

```
$ sudo howdy config

Opening config.ini in the default editor
Traceback (most recent call last):
  File "/usr/lib/python3.13/site-packages/howdy/cli.py", line 105, in <module>
    import cli.config
  File "/usr/lib/python3.13/site-packages/howdy/cli/config.py", line 23, in <module>
    subprocess.call([editor, paths_factory.config_file_path()])
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 397, in call
    with Popen(*popenargs, **kwargs) as p:
         ~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 1038, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                        pass_fds, cwd, env,
                        ^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
                        gid, gids, uid, umask,
                        ^^^^^^^^^^^^^^^^^^^^^^
                        start_new_session, process_group)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 1859, in _execute_child
    self._posix_spawn(args, executable, env, restore_signals, close_fds,
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      p2cread, p2cwrite,
                      ^^^^^^^^^^^^^^^^^^
                      c2pread, c2pwrite,
                      ^^^^^^^^^^^^^^^^^^
                      errread, errwrite)
                      ^^^^^^^^^^^^^^^^^^
[...]

Issue #983, opened 2025-01-09
Better $EDITOR handling


aleeraser:
The current EDITOR handling is a bit lacking.

https://github.com/boltgolt/howdy/blob/aa75c7666c040c6a7c83cd92b9b81a6fea4ce97c/howdy/src/cli/config.py#L14-L20

Improvements:
* check that `/bin/nano` exists before trying to spawn a subprocess
  * or better, search for `nano` in PATH instead of using a hardcoded path
* use a universally-included editor like `vi` as default
  * if the goal was to use a more beginner-friendly editor, then perhaps resort to `vim` only if `nano` is not available
* suggest using `sudo -E howdy [...]` in the documentation, as otherwise the command would run in a non-interactive shell (i.e. without sourcing the initialization files), and no environment variable is carried over (including EDITOR) unless explicitly preserved

boltgolt:
Sounds good, please open a PR

aleeraser:
I took a quick look and couldn't see this completed. Can you point out where was this addressed?  
Also, sorry for not having the time to properly address this.

boltgolt:
No worries, lack of time has been my issue too. Just closed this as stale

PR #1009, opened 2025-04-01
feat(config): better EDITOR handling

aleeraser:
The current EDITOR handling is a bit lacking.

Improvements:

- check that the editor exists before trying to spawn a subprocess
- add vi as nano alternative
- if everything else fails, perhaps the user is using sudo. Suggest using 'sudo -E howdy [...]', as otherwise the command would run in a non-interactive shell (i.e. without sourcing the initialization files), and no environment variable is carried over (including EDITOR) unless explicitly preserved

Related:
- https://github.com/boltgolt/howdy/issues/983
- https://github.com/boltgolt/howdy/issues/1000

### DIFF
--- a/howdy/src/cli/config.py
+++ b/howdy/src/cli/config.py
@@ -3,21 +3,40 @@
 # Import required modules
 import os
 import subprocess
+import shutil
 import paths_factory
 
 from i18n import _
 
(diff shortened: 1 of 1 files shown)


boltgolt:
Thank you!

PR #1032, opened 2025-07-17
Add Intel IPU6 and libcamera support via GStreamer integration

bilalobe:
# Add optional GStreamer-assisted capture path for systems where direct OpenCV capture is not usable

## Summary

This PR adds an optional way for Howdy to start a user-configured GStreamer command before opening the configured video device.

The goal is narrow: support systems where the camera cannot be read reliably through the existing direct OpenCV path alone, but can be made available through a user-managed capture pipeline.

## What this changes

This PR adds three optional configuration values in the `[video]` section:

- `use_gstreamer`
- `gstreamer_pipeline`
- `gstreamer_startup_delay`

When `use_gstreamer = true`, Howdy will:

1. start the configured GStreamer command as a subprocess
2. wait for the configured startup delay
3. open `device_path` using the existing recorder flow
4. terminate the subprocess when capture is released

If `use_gstreamer = false`, behavior remains unchanged.

## What this does not do

This PR does not:
- replace the current capture implementation
- add native libcamera integration to Howdy
- auto-detect pipelines or camera backends
- change the default configuration for existing users

The configured `device_path` is still the device Howdy reads from. This change only adds an optional subprocess startup step before capture.

## Why this is useful

On some systems, especially newer laptop camera stacks, the configured capture device may need to be created or prepared externally before OpenCV can use it.

This PR allows that preparation step to be driven by configuration instead of requiring local wrapper scripts or manual process management outside Howdy.

## Behavior and failure mode

- disabled by default
- existing setups are unaffected unless explicitly enabled
- if enabled without a configured pipeline, Howdy exits with an error
- if the configured command cannot be started, Howdy exits with an error
- subprocess shutdown is handled during release/destruction

## Files changed

- `howdy/src/recorders/video_capture.py`
- `howdy/src/config.ini`
- `howdy/src/GSTREAMER_INTEGRATION.md`

### DIFF
--- a/howdy/src/GSTREAMER_INTEGRATION.md
+++ b/howdy/src/GSTREAMER_INTEGRATION.md
@@ -1,0 +1,111 @@
+### **`GSTREAMER_INTEGRATION.md`**
+
+# Enabling Modern Camera Support (libcamera / IPU6) with GStreamer
+
+This guide explains how to use Howdy's GStreamer integration feature. If you have a modern laptop (roughly 2020 or newer) and your built-in camera does not work with Howdy out of the box, you are in the right place.
+
+## The Problem: Modern Cameras vs. Older Applications
+
+Many new cameras, especially the Infrared (IR) cameras used for facial recognition (like those compatible with Windows Hello), no longer present themselves as simple webcams. They use a modern driver stack called `libcamera`.
+
+Howdy, by default, uses a library (OpenCV) that expects a standard V4L2 video device (`/dev/videoX`). The `libcamera` stack does not provide a stable device in this format, causing Howdy to fail with "camera not found" or "failed to read frame" errors.
+
+## The Solution: An On-Demand GStreamer Bridge
+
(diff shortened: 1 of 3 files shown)


afalout:
Hello, 

first off, a big Thank You @bilalobe for your effort.

This functionality is very much needed, and will only be more necessary going forward. I have several Microsoft Surface tablets and laptops and would love to have this working.

@boltgolt - did you have a chance to take a look at this PR?

Thanks,
Andrej

alba4k:
this looks quite vibe-coded, and I'm not just referring to the PR body.

But if it works as promised, this is great :)

Although I have had howdy working on my xps 9320 even without this

bilalobe (commented):
I revisited this on my own IPU6/libcamera machine and want to tighten the framing a bit.

  The core problem is real: on newer laptops, the camera stack often does not present a stable desktop-friendly V4L2 stream that OpenCV can consume directly. In practice, I ended up needing a bridge of the form:

  `libcamera / libcamerasrc -> GStreamer -> v4l2loopback -> OpenCV/Howdy`

  So the underlying motivation for this PR is still valid.

PR #1049, opened 2025-10-13
howdy-gtk: fix python path finding

fufexan:
Fix trying to get the python path from the system instead of the meson option.


AkechiShiro:
Hi @boltgolt,

Sorry to ping again on this PR but it would be quite nice if they could be merged, this would improve the package for Nix/NixOS based distros.

There is also #1050

If you have any concerns, feel free to share them.

Kind regards,

fufexan:
This is no longer needed. My issue stemmed from a double-wrapped python, which I've solved in the meantime.

For more context: https://github.com/NixOS/nixpkgs/pull/216245#discussion_r2681317586.

fufexan:
Nevermind, my testing was wrong. It is still needed.

PR #1064, opened 2025-11-20
Don't overwrite config.ini on reinstall

nafg:
When I installed from source it overwrote my config. This seems to fix that.

(I used Claude Code for this)

The fs.exists() check only runs during meson setup, not during install. This meant that once the build was configured, every 'meson install' would overwrite the user's config.ini file.

Comment out the config.ini installation to preserve user configurations during reinstalls when building from source.


nafg:
:facepalm:

PR #1128, opened 2026-07-10
fix: make test command histogram logic shape-agnostic for OpenCV 5

### LINKED ISSUE #1127: Howdy test results in IndexError: invalid index to scalar variable
Hi,

I installed last howdy version from github following instructions for Fedora:
- 1st try: https://github.com/boltgolt/howdy/issues/1069#issuecomment-3592401752
- 2nd try: https://github.com/valeriopastore20/howdy-guide

Each time I can launch "sudo howdy add" and it seems successful, the camera lighting indicator is activated during the process. But nothing works in GDM.

When I try "sudo howdy test" I get the following:
`QFontDatabase: Cannot find font directory /home/olivier/.local/.howdyvenv/lib64/python3.14/site-packages/
cv2/qt/fonts.
Note that Qt no longer ships fonts. Deploy some (from https://dejavu-fonts.github.io/ for example) or switch to fontconfig.
QFontDatabase: Cannot find font directory /home/olivier/.local/.howdyvenv/lib64/python3.14/site-packages/cv2/qt/fonts.
Note that Qt no longer ships fonts. Deploy some (from https://dejavu-fonts.github.io/ for example) or switch to fontconfig.
QFontDatabase: Cannot find font directory /home/olivier/.local/.howdyvenv/lib64/python3.14/site-packages/cv2/qt/fonts.
Note that Qt no longer ships fonts. Deploy some (from https://dejavu-fonts.github.io/ for example) or switch to fontconfig.
QFontDatabase: Cannot find font directory /home/olivier/.local/.howdyvenv/lib64/python3.14/site-packages/cv2/qt/fonts.
Note that Qt no longer ships fonts. Deploy some (from https://dejavu-fonts.github.io/ for example) or switch to fontconfig.
QFontDatabase: Cannot find font directory /home/olivier/.local/.howdyvenv/lib64/python3.14/site-packages/cv2/qt/fonts.
Note that Qt no longer ships fonts. Deploy some (from https://dejavu-fonts.github.io/ for example) or switch to fontconfig.
  File "/usr/local/lib64/howdy/cli.py", line 116, in <module>
    import cli.test
  File "/usr/local/lib64/howdy/cli/test.py", line 131, in <module>
    hist_total = int(sum(hist)[0])
                     ~~~~~~~~~^^^
IndexError: invalid index to scalar variable.`

I don't care about the QFontDatabase error, but what about cli.py and test.py ?

faux123:
## Bug

`sudo howdy test` crashes on systems with OpenCV 5.x:

```
  File "/usr/lib/howdy/cli.py", line 116, in <module>
    import cli.test
  File "/usr/lib/howdy/cli/test.py", line 131, in <module>
    hist_total = int(sum(hist)[0])
                     ~~~~~~~~~^^^
IndexError: invalid index to scalar variable.
```

Fixes #1127

## Root cause

OpenCV 5.0 changed the Python binding of `cv2.calcHist` to return a 1-D array. With the 8-bin call in `test.py` it now returns shape `(8,)` where OpenCV 4.x returned `(8, 1)`:

```python
>>> import cv2, numpy
>>> cv2.__version__, numpy.__version__
('5.0.0', '2.5.1')
>>> f = numpy.random.randint(0, 255, (120, 160), dtype=numpy.uint8)
>>> cv2.calcHist([f], [0], None, [8], [0, 256]).shape
(8,)
```

With a 1-D array, built-in `sum(hist)` produces a `numpy.float32` scalar instead of a length-1 array, so the `[0]` index raises `IndexError`. The subsequent `value[0]` in the bar-drawing loop fails the same way.

## Fix

Flatten the histogram with `ravel()` and total it with `hist.sum()`, so the logic works with the array layout of every OpenCV version (4.x `(8, 1)`, 5.x `(8,)`) instead of assuming either one. `compare.py` and `add.py` are unaffected: they already use `np.sum(hist)` and `hist[0]`, which work with both shapes.

## Environment where reproduced

- CachyOS (Arch-based), howdy master (d3ab993, via AUR `howdy-git`)
- opencv 5.0.0, numpy 2.5.1, Python 3.14.6, dlib 20.0.1
- Reporter in #1127 hit the identical traceback on Fedora with Python 3.14

## Testing

- Patched logic exercised against real `cv2.calcHist` output under OpenCV 5.0.0, and synthetically against both the `(8, 1)` and `(8,)` layouts.
- Full `howdy test` loop (this exact patched file) run against a real IR camera for 30 frames on the system above: histogram bars computed, no errors, clean exit.
- Interactive `sudo howdy test` confirmed working with the equivalent fix applied to the installed copy.


Raymo111:
@boltgolt can we please merge

